PR #3, opened 2021-10-19
Various changes

CerberusHellHound:
Modifies Internal and External organs
- Headshot can be deadly (as brain is now fragile, while keeping the initial 200 health allowing for eventual revival in due time; a headshot can still fuck someone up badly if it hits the brain)
- Internal organs are slightly nerfed, so that getting shot in the chest will be a bad experience (not to the extent that it was on the temp code) 
- Started working on projectile defs (not done at all.. we'll need to add the new sprites too)
- Limbs are tough, hard to decap or delimb

### DIFF
--- a/code/modules/organs/external/head.dm
+++ b/code/modules/organs/external/head.dm
@@ -3,8 +3,8 @@
 	icon_name = "head"
 	name = "head"
 	slot_flags = SLOT_BELT
-	max_damage = 75
-	min_broken_damage = 35
+	max_damage = 175
+	min_broken_damage = 25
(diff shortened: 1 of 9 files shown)


billyjoebobman:
this pr suxs, doesnt account for suffocation so if you get strangled for like 20 seconds you die, if you get hit in the chest a few times with a telescopic baton you end up with major organ damage, even with decent bulletproof armor if you get shot a few times in the chest you are likely just going to die from organ damage where with the old system it would just be pain, the old system generally was way better in terms of melee and gun combat and some things just dont seem like they work, with one time I shot someone in the head and they just walked off. pls fix

PR #83, opened 2022-06-25
Adds more AI names

Calyxman:
Adds NATO phonetic AI names, AIAD AI names (as well as some not on the wiki), keeps some old AI names, as well as some new ones based off Greek gods and other such things.


Calyxman:
Poke me over Discord if this doesn't work. I haven't used github in a while.

PR #104, opened 2022-06-26
Map-Hotfixes (53-2, 53-3, 53-4)

TerryTehCoder:
Removes a bunch of Gyroton Fusion Injectors that popped up as a result of merge conflicts.

Replaces machinery that was eaten by the fusion injectors.

Re-adds server farm security cameras.

Second instance of 131-A deleted.

Random airlock near server farm that was left in from someones previous mapping PR was deleted.

Take 2, this time hopefully we won't have an entire TGM conversion in the way.


Calyxman:
Wait, why did we delete 131-B?

PR #108, opened 2022-06-27
Makes Residents, Factions lore accurate, more AIC names

Calyxman:
Adds some new AIC names, some are based off NATO phonetic alphabet.
Updates location_human.dm properly - no more Bay planets
Deletes the other corps.
Poke me if this doesn't work, or isn't lore-friendly.


Bierkraan:
Pretty sure this will cause a compile error, did you test this?

PR #109, opened 2022-06-27
Give surgeon proper access

T-rexsuper1:
For some reason surgeon does not have the access he's supposed to have


T-rexsuper1:
other PR will handle this issue

PR #161, opened 2022-07-04
UIU logo, fax office targets for both GoI reps and access fix for them

StiffRobot:


tichys:
Removes so me closes, does this compile and work on your end through testing?

PR #164, opened 2022-07-04
Organic value changes (more lethal combat)

CerberusHellHound:
Headshots without helmet are now very deadly.

Bodyshots are now likely to be very deadly, but less so than headshots.


tichys:
Will require heavy balancing with armor

PR #179, opened 2022-07-06
Renaming, slight rebalance, etc

CerberusHellHound:
List of changes : 
- decreased brain health to 150 (instead of 200), it's high enough that medical assistance can be given if fast but low enough that you don't want to get shot
- increased damage values of weapons to baystation/nebula level (40 for a pistol for eg) 
- increased adrenalin generation when hurt (less fading in and out, you can still use your gun when hit and pain won't be such a pain in the ass, but you're less likely to get back up once the final shot hits you)
- decreased relative size of lungs from 60% to 30% so that now, getting hit in the chest won't have as much chance of damaging your poor fucking lungs (yes 60% is the original baystation number.. it makes sense, it's a large organ, but it's a pain in the ass)
- changed some names and descriptions of certain weapons and firearms to better fit established naming convention
-made revolvers cycle the barrel instead of eject each shot (it's a revolver, not a damn rifle) 
- slightly decreased firing delay for the mk9 revolver (slightly weaker than the mateba so slightly faster firing) 
- decreased firing delay for the mk9 pistol (lower caliber, less recoil, easier to magdump)

Todo next : 
I'll be working on armour values. They aren't too bad at the moment but I'll try to find a nice balance.


EgorDinamit:
I think it's not the best idea to make combat even more deadly.
It'd be much better for there to be longer fire-fights that don't end with 1 shot.

PR #184, opened 2022-07-07
Cryopods allow instant respawn

CocaColaTastesGood:
Changes:
Ghosting while in a cryopod(Or waiting for your body to despawn) will allow instant respawn. 
Notes:
This may be abused, but I think it's better to handle things like that OOC. This only affects the human cryopod.


CocaColaTastesGood:
I can make it so it only removes part of the timer.

PR #186, opened 2022-07-07
Adds a blacklist to 049s targetting/343 immune to 049

CocaColaTastesGood:
Changes:
049 has a list of mob types that it cannot target. The only thing in the last as of now is 343.


CocaColaTastesGood:
typo

PR #239, opened 2022-07-09
Sets chance for psionic latency to 2%

EgorDinamit:
### About pull request

Simple as that. Chance for psionic latencies(not powers themselves, mind you) is set to 2%.
Effectively this means that 1 out of 50 people will have a chance to awaken random psionic powers if they:
- A: Receive near deadly amount of brain damage (40% chance).
- B: Overdose on Three-Eye drug (tiny chance and requires overdose, which is quite deadly).
- C: Use Jerraman drug (almost guaranteed chance, but hard to make. DO NOT take more than 4 units).

### DIFF
--- a/code/game/jobs/job/job.dm
+++ b/code/game/jobs/job/job.dm
@@ -50,8 +50,8 @@
 	var/defer_roundstart_spawn = FALSE // If true, the job will be put off until all other jobs have been populated.
 	var/list/species_branch_rank_cache_ = list()
 	var/list/psi_faculties                // Starting psi faculties, if any.
-	var/psi_latency_chance = 0            // Chance of an additional psi latency, if any.
-	var/give_psionic_implant_on_join = TRUE // If psionic, will be implanted for control.
+	var/psi_latency_chance = 2            // Chance of an additional psi latency, if any.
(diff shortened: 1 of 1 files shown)


garrulousTeamotei:
Is there a means to obtain carpotoxin to create the Three-Eye drug?

EgorDinamit:
> Is there a means to obtain carpotoxin to create the Three-Eye drug?

That's a very good question.
We can consider adding ways to activate psionics later.

garrulousTeamotei:
Could swap out carpotoxin for something similar. From time consuming mob hunting to time consuming botany. Maybe Amatoxin (in higher amounts to reward obtaining it from Destroying Angel instead of Amanita. That, or Cyanide from Poison Apple, or Poison Berry Juice from Poison Berries. All toxins, to fit in with the Carpotoxin theme, and none available by normal chemistry.

CocaColaTastesGood:
Crystallising agent doesn't activate psionics. You can buy a russian seafood vendor from logi and hack it for carpotoxin. You can also emag the cargo computer to buy three-eye in a contraband crate. Everything has a probability of working. Everything except for the luxury version of three-eye has a pretty high chance of killing you outright though.

PR #241, opened 2022-07-10
Engistuff removal, EZ holding addition

Tsurupeta:
Removed RIGs and hardsuits from engis, CE and CE. Added them 2 rad and 2 firehazard lockers instead.

Added 3 holding cells to the EZ.

Also moved the ballistic vests and helmets to locker spawn.

image


Vizoplez:
>doesn't remove the storage for it

Tsurupeta:
I removed the storage for it, but because of some mapfuckery afterwards it was half-reverted, so one of the suits was brought back and even more shit broke.

PR #247, opened 2022-07-10
SCP-106 Rework

### LINKED ISSUE #208: bug report: SCP 106 can feel pain, but 49 can't
#### Description of issue
SCP 106 can feel pain,m but 49 can't

SCP 106 ignoring pain and killing people while 49 cowers in fear after being slapped

frail old man vs doctor high in his own medication
#### Steps to reproduce
code bug my man

#### Issue bingo

- [ ] Issue could be reproduced at least once
- [ ] Issue could be reproduced in multiple rounds
- [ ] Issue happened in a recent (less than 7 days ago) round
- [ ] [Couldn't find an existing issue about this](https://github.com/Baystation12/Baystation12/issues)

ItsTisme:
Rework how SCP-106 works. Giving him ability to go into  a wall and re-appear from the same wall else where. Making him EXTREMELY mobile

- [x] Bassic wall phasing system
- [x] Make his bones unbreakable/amputatable
- [x] Add delay to powers
- [ ] Add a stun gun for SCP 106(optional)
- [ ] Rework how he grabs people(optional)
- [ ] Fix bugs(never done)

Ill do a part 2 of this later if needed since the TM is merged anyway and it will prop cause confusion

closes  #208

gif

:cl: Autisme
tweak: Reworked SCP-106
/:cl:


Tsurupeta (commented):
quick review, didn't look at code too deeply

code/modules/mob/living/life.dm
```diff
@@ -195,7 +195,7 @@
 /mob/living/proc/update_sight()
 	set_sight(0)
 	set_see_in_dark(0)
-	if(stat == DEAD || eyeobj)
+	if(stat == DEAD)
```
Autisem, line 198: Can some tell me why this even is a thing? it litervly fires on every tick of life and it seems quiet dumb

Autisem: I figured it out and I hate that this is how it works

EgorDinamit:
Fancy, but I'd recommend to avoid using verbs all-together.
It'd be nice if we could put that function into simply clicking or bumping into a wall.

Autisem:
> Fancy, but I'd recommend to avoid using verbs all-together. It'd be nice if we could put that function into simply clicking or bumping into a wall.

I want to avoid it being trigger accidentally a bunch. all the SCP related verbs are under there which is why it didnt alarm me to much

Tsurupeta (approved):

PR #254, opened 2022-07-10
GOC base, 008 fix, map fix, a lot more shit

Bierkraan:


CerberusHellHound (changes requested):
Current gun balance feels good, imo, you nerfed it to hell and back

PR #285, opened 2022-07-11
I HATE MODULARITY I HATE IT (096 update)

Yawet330:
Makes 096 aggro through pictures.
Very scary. Very dangerous.

### DIFF
--- a/code/modules/SCP/SCPs/SCP-096.dm
+++ b/code/modules/SCP/SCPs/SCP-096.dm
@@ -1,9 +1,7 @@
+//sculpture
 //SCP-096, nothing more need be said
 //Do this here as it'll be the only one that uses this - Yawet330 (for now)
-/mob/living/simple_animal/scp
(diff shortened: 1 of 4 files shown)


code/modules/SCP/SCPs/SCP-096.dm
```diff
@@ -1,5 +1,6 @@
 //sculpture
 //SCP-096, nothing more need be said
+//Do this here as it'll be the only one that uses this - Yawet330 (for now)
```
Penwin0, line 3: I don't think this comment is very clear on whatever it's trying to convey

CocaColaTastesGood:
> For reference you can pin pictures of 096 to d-block cell doors and if anyone examines it, he'll aggro across z-levels and gib them on the spot, resulting in funny mass-death.
> 
> May actually need balance considerations

Biggest issue I think is that he intentionally goes out of his way to be seen by following players.

PR #286, opened 2022-07-11
I HATE TICHYS!!

Yawet330:
fixes him overwriting one of my previous PRs (antag droon)


PrimeGenerations:
That's if the drones are not in maintenance secretly chopping away at the facility. I don't think this is a good PR IMHO

PR #300, opened 2022-07-12
Removes AKs from LCZ. Stop adding AKs

Teragenic:
Removes AKs from LCZ, update your damn maps.


PrimeGenerations:
you can close your PR, I updated this with the third or fourth attempt of implementation of the Security Update

PR #301, opened 2022-07-12
Modifies runechat to not show whispers across the room, and adds whis…

Vivalas:
…per verbs to all current languages to make it more clear that people are whispering.

As title says. Runechat was overriding default behavior of whispers, this simply updates runechat to use that behavior and adds whisper verbs to all spoken languages.


EgorDinamit:
God bless.

PR #308, opened 2022-07-12
Fix species runtime

Autisem:
So since new happens before initalize, and defaults to human specices and human organs. on init it will have to re-set the specices qdelling old organs which causes runtimes. The 106 fix ill be in my rework PR

### DIFF
--- a/code/modules/SCP/SCPs/SCP-343.dm
+++ b/code/modules/SCP/SCPs/SCP-343.dm
@@ -16,6 +16,10 @@ GLOBAL_LIST_EMPTY(scp343s)
 /mob/living/carbon/human/scp343/IsAdvancedToolUser()
 	return FALSE
 
+/mob/living/carbon/human/scp343/New(new_loc, new_species)
+	new_species = "SCP-343"
+	. = ..()
+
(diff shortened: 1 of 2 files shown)


code/modules/SCP/SCPs/SCP-343.dm
```diff
@@ -16,6 +16,10 @@ GLOBAL_LIST_EMPTY(scp343s)
 /mob/living/carbon/human/scp343/IsAdvancedToolUser()
 	return FALSE
 
+/mob/living/carbon/human/scp343/New(new_loc, new_species)
+	new_species = "SCP-343"
+	. = ..()
```
ivanmixo, line 21:
```suggestion
	return ..()
```

baystation12.dme
```diff
@@ -125,6 +125,7 @@
 #include "code\_helpers\medical_scans.dm"
 #include "code\_helpers\mobs.dm"
 #include "code\_helpers\names.dm"
+#include "code\_helpers\pronouns.dm"
```
Autisem, line 128: not sure if this is needed still

PR #309, opened 2022-07-12
Changed SCP-500 quantity of Adminordrazine to 1u from 50

Magisterium2022:
See title.


Penwin0:
Diff seems to be fine, so it shouldn't be a problem, a bit ugly though

PR #328, opened 2022-07-13
Minor chemist skill buff

breadcrombs:
It didn't make sense that chemists are not experienced in chemistry by default but the surgeons are.


breadcrombs:
I meant to say that surgeons are experienced in surgery ( as both experienced medicine and anatomy) but chemists aren't experienced in chemistry (only basic in chem).

breadcrombs:
oops forgot to name commit

PR #332, opened 2022-07-13
minor chemist skill buff

breadcrombs:
changes chemist's chemistry skill from basic to experienced in chemistry.


Vivalas:
This isn't actually a buff, it just determines how many points must be allocated by default to chemistry. That being said, it's not a bad idea

PR #373, opened 2022-07-13
Give SD's locker a very light concealable armor vest.

breadcrombs:
Gives SD the option to have a slight degree of protection while wearing some outer formalwear. The concealable armor only grants MINOR protection and only covers the upper and lower area. Functions just like a light plate but you can put it as a shirt accessorie.


breadcrombs:
the 2 bots are still broken

PR #377, opened 2022-07-14
Smes fix

Vivalas:
added logging for SMES. don't think this is an exploit but it should be here anyways, since cutting all the SMES wires can break every APC on the entire map


Vivalas:
don't merge, need to rebase out that commit

PR #378, opened 2022-07-14
logging added for SMES grounding

Vivalas:
don't think this is an exploit, but logging is still nice to have, since this can destroy every single apc on the map


Vivalas:
check failed is because I used non-static typing (: operator), but as I explained in code there's not really a better way

PR #379, opened 2022-07-14
Fix APC's calling area power_change each tick

checkraisefold:
## About the Pull Request

Fixes a bug in APCs that was being used to crash the server

## Why It's Good For The Game

no more crash on my 150 player server plz

## Did you test it?

yes, the power outage event works normally. everything works fine

## Changelog

:cl:
bugfix: Fixes an APC server lagging bug.
/:cl:


Vivalas:
Hold off on this, I have a better implementation in the works

code/modules/power/apc.dm
```diff
@@ -799,8 +800,6 @@
 		area.power_equip = 0
 		area.power_environ = 0
 
-	area.power_change()
```
Vivalas, line 802: this completely breaks APC update functionality, this is not a good fix

Vivalas (changes requested):

Vivalas (approved):
looks good

PR #396, opened 2022-07-14
Adds concealable armor plate into the SD's locker

breadcrombs:
Site director's now get a light armor plate that can be worn as a accessories over the shirt. No longer needs to steal armor or give up that formal outerwear.


breadcrombs:
one line code change and now 3 bots are broken instead of 2

Vivalas:
code/map bots are spergs atm

PR #405, opened 2022-07-14
Vision cone perfomance hotfix

Tsurupeta:
Removing the item deletion from vision cone, until full VC rework.

GC can't handle it.


Autisem (approved):
lets see how this goes

PR #435, opened 2022-07-15
Fixes game log time. Now it uses real world time

DTraitor:


Vivalas (approved):
good to merge after checks

PR #451, opened 2022-07-16
fixes more stuff I broke

DTraitor:
## About the Pull Request

* Fixed alt click menu.
* Debugger tweaks
* Robot "Subsystems" tab is back. It's not really optimized because it sends data each 4 ticks. Someone should change it so it will send new data only when robot submodules is changed.


AC2ServerAccount:
Based man, thank you.

PR #472, opened 2022-07-18
Loadout tweak (Restrict some nonsensical gear option from Security roles while keeping sensical ones)

CerberusHellHound:
## Why It's Good For The Game
Reason 1) Decided by the community, see attached poll screenshot
Reason 2) D-Class role already has heavy restriction on loadout options. Mostly because they are prisoners/test subjects who should have some amount of uniformity to themselves. 
So why shouldn't security officers belonging to an highly militarized department of the Foundation? 
(fixed now)_Reason 3) Some slots would overwrite existing outfits for security roles (armour, uniform, etc)._ (fixed now)
Reason 4) The restrictions on loadout options are far from complete. Most options are present, but there are restrictions only on extremely nonsensical items and stuff that would completely break uniformization. If its not highly visible and flashy, it's probably not on the list of restricted items.

## Changelog

:cl:
- Gears that would not be realistically worn by sec are mostly restricted
- All utilities (tablets, etc), the vast majority of accessories (necklaces, rings etc) and several types of headgear (beanie, beret...) are unrestricted
- Several types of boots (workboot, jackboots variants etc) are allowed
- Work gloves are allowed
- Uniform options are completely restricted, so are exowear/suit options (none of the options listed in the loadout menu makes sense, and they eat the slot) in accordance with the pol (keeping uniformity, these options are huge visual differences)
- Tactical section is largely still accessible as well, tho keep in mind many of the available headcovers reference NT/Solgov stuff
:cl:


CerberusHellHound:
image

Vivalas:
Just fyi, reason 3 isn't valid anymore, since I fixed loadouts overwriting gear.

code/modules/client/preference_setup/loadout/lists/headwear.dm
```diff
@@ -43,6 +46,7 @@
 	display_name = "mariner's cap, colour select"
 	path = /obj/item/clothing/head/mariner
 	flags = GEAR_HAS_COLOR_SELECTION
+	allowed_branches = list(/datum/mil_branch/security)
 
 /datum/gear/head/orange_cap
 	display_name = "orange cap"
```
Vivalas, line 52: Re-add /datum/gear/head/cap below, please. Patrol caps are a basic military/security uniform item. If you want, split it into two items, a "uniform cap selection" that removes the purple/rainbow etc. and then the rest.

code/modules/client/preference_setup/loadout/lists/suits.dm
```diff
@@ -3,18 +3,20 @@
 	sort_category = "Suits and Overwear"
 	category = /datum/gear/suit
 	denied_roles = list(/datum/job/classd)
-	//Reasoning : it eats the Armour slot of guards. Also, there's little justification for wearing one of these as a guard
+	allowed_branches = list(/datum/mil_branch/civilian)
+	//Reasoning : it eats the Armour slot of guards. None of these options fits security. Too flashy, too visible, uniformity would be broken. To note: some guards have access to vest options at their lockers.
```
Vivalas: It doesn't. not anymore.

code/modules/client/preference_setup/loadout/lists/suits.dm
```diff
@@ -3,18 +3,20 @@
...
 
 /datum/gear/suit/poncho
 	display_name = "poncho selection"
 	path = /obj/item/clothing/suit/poncho/colored
 	cost = 1
 	flags = GEAR_HAS_TYPE_SELECTION
 
-/datum/gear/suit/security_poncho
+//no poncho for foundation guards, sorry
```
Vivalas: Why are you adding a branch up above then commenting this anyways?

code/modules/client/preference_setup/loadout/lists/suits.dm
```diff
@@ -27,10 +29,11 @@
...
-/datum/gear/suit/nanotrasen_poncho
+//NT descriptions
+/*/datum/gear/suit/nanotrasen_poncho
 	display_name = "poncho, NanoTrasen"
 	path = /obj/item/clothing/suit/poncho/roles/science/nanotrasen
-
+*/
 /datum/gear/suit/cargo_poncho
 	display_name = "poncho, supply"
 	path = /obj/item/clothing/suit/poncho/roles/cargo
```
Vivalas, line 38: Can't comment below but leather coat / trenchcoat would be valid for command security roles, imo, where uniform restrictions would be a mite more relaxed for higher ranking personnel, and also disgingush them, can change it or I may just go add it to their uniform or just enable it for certain roles.

Vivalas (changes requested):

code/modules/client/preference_setup/loadout/lists/suits.dm
CerberusHellHound: Fair enough, removed the comment

code/modules/client/preference_setup/loadout/lists/headwear.dm
CerberusHellHound: Fair enough, willdo
image

code/modules/client/preference_setup/loadout/lists/suits.dm
CerberusHellHound: Alright, indicated in description

code/modules/client/preference_setup/loadout/lists/suits.dm
CerberusHellHound: Good point, i'll unrestrict it for GC and zonecomms...

CerberusHellHound: image

Vivalas (approved):

PR #496, opened 2022-07-19
Makes robots a little bit more neutral.

AC2ServerAccount:
## About the Pull Request

Adds a 5th law  "An AIC must take a neutral stance on conflicts unless one side is at risk of causing destruction of the foundation.""

Took a bit to discuss a wording that would work but I feel this leaves enough open to make robots have the choice in conflicts but also disallowing them to engage in constant small scuffles with D-Class. 

For example an interpretation could be that D-Class rioting in just DCZ is not at risk of causing destruction of the foundation. But D class that have escaped DCZ and are almost at the surface are at risk of causing destruction to the foundation due to a possible info breach.

 The lawset also does not specifically state what stance must be taken after said level is reached so unless the players interpretation of law 3("An AIC must operate for the benefit of the Foundation.") is violated they could remain neutral. 

## Why It's Good For The Game

I know some people have seemed upset at robot players recently and while I don't totally agree on a lot of things there is definitely some issues with the way some characters play and I think this will make both sides happy, 

## Changelog

:cl:
add: Law 5  "An AIC must take a neutral stance on conflicts unless one side is at risk of causing destruction of the foundation."
/:cl:

### DIFF
--- a/code/datums/ai_law_sets.dm
+++ b/code/datums/ai_law_sets.dm
@@ -156,7 +156,7 @@
 /datum/ai_laws/foundation/New()
 	src.add_inherent_law("An AIC must know it is an AIC. You are an AIC.")
 	src.add_inherent_law("An AIC must not operate outside of its clearance.")
-	src.add_inherent_law("An AIC must operate for the benefit of the Foundation.")
+	src.add_inherent_law("An AIC must operate for the benefit of the Foundation, except in situations that would violate the 5th law.")
 	src.add_inherent_law("An AIC must protect its own existence unless it conflicts with other principles.")
 	src.add_inherent_law("An AIC must take a neutral stance on conflicts unless one side is at risk of causing destruction of the foundation.")
 	..()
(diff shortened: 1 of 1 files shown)


ivanmixo (approved):
Code is okay.

Vizoplez:
No, this would contradict the third law.

"The AIC must act to the benefit of the foundation."

From a lore and gameplay perspective almost every situation where conflict occurs, the AIC will 9/10 times have to support foundation-related activities, meaning law 5 will be useless, and generate nothing but law conflict.

AC2ServerAccount:
> No, this would contradict the third law.
> 
> "The AIC must act to the benefit of the foundation."
> 
> From a lore and gameplay perspective almost every situation where conflict occurs, the AIC will 9/10 times have to support foundation-related activities, meaning law 5 will be useless, and generate nothing but law conflict.

I added a clause to 3 to fix this. As stated in the description this does not force their stance after the point at which a law 5 violation is reached and law 3 is heavily open to interpretation. This is mainly to prevent people who go around playing security and smacking people during every small D-Class scuffle.

AC2ServerAccount:
If your going to emote, please leave a comment why, I am perfectly willing to change the wording if you have concerns

PR #503, opened 2022-07-20
HCZ Hazmat Suit

dawsonkeyes:
Gives the Heavy Containment Zone guard team a unique hazmat suit with no slowdown that acts as both capable armor (equivalent to their old gear) and protection from biological and fire-based threats. No longer shall man be shackled by shitty berets and balaclavas!

image


ItsVyzo:
unit tests, resolve them

dawsonkeyes:
VIVAAAAAAAAAAAAAAA

Vivalas:
Pretty sure this ain't mapmerged

Vivalas:
Also, uh, why did you remove the hazmat suit from the locker instead of just adding it to locker contents? That's the preferred method to mapping it in. If that's what was causing the pre/post-init checks it means you're stuffing too much crap in and you need to listen to the unit check.

User-U-U:
do you need to working version of map merge one we have rn isn't working I can upload it tmrw

dawsonkeyes:
> Also, uh, why did you remove the hazmat suit from the locker instead of just adding it to locker contents? That's the preferred method to mapping it in. If that's what was causing the pre/post-init checks it means you're stuffing too much crap in and you need to listen to the unit check.

I removed the existing biosuit in the lockers it was added into as they served the same function, and also because of locker overstuffing

Vivalas:
If lockers are overstuffed, you can't just paste it on the map. I'm fairly certain it will just pop back out. If not, still, the preferred method is to have it in the locker. My recommendation is to make a new HCZ biosuit locker and place it into their equipment rooms.

Vivalas:
Either way mapmerge needs to be fixed

ivanmixo (changes requested):
Look I'm sure you did 150k mapping changes just to add some hazmat suits but you seriously need a map merger my guy.

Nigkgro99:
wouldn't the easy way of getting around this just be to put ALL of the armor on a rack or in one normal locker? it'd be lazy but it wouldn't involve all of this stupid bullshit

dawsonkeyes:
i have a plan

PR #515, opened 2022-07-21
GOC Updates (Large BUFF to GOC-ERT)

AMcilraith:
This makes the GOC more in line with their canon, giving them better armour and more weapons.

## About the Pull Request

Adds hazard suits to the GOC Base.
Adds more grenades to GOC Base in the stead of ammo.
Adds a proper chem dispenser to the GOC Base.

## Why It's Good For The Game

It's good for the gsme because:
A) GOC is rarely used
B) GOC is now more in line with the canon

## Changelog

:cl:
add: hardsuit in goc base
add: ert chem dispenser in goc base
balance: rebalanced goc, gives buffm makes more in line with canon

/:cl:

### DIFF
--- a/maps/site53/z1_admin.dmm
+++ b/maps/site53/z1_admin.dmm
@@ -1406,45 +1406,6 @@
 /obj/machinery/light,
 /turf/simulated/floor/tiled/techmaint,
 /area/site53/lhcz/scp1102entrance)
-"iz" = (
-/obj/item/clothing/head/helmet/scp/goc,
-/obj/item/clothing/suit/armor/goc,
-/obj/structure/closet,
-/obj/item/storage/belt/holster/security,
-/obj/item/clothing/gloves/thick/swat,
-/obj/item/clothing/shoes/jackboots,
-/obj/item/clothing/under/ert,
-/obj/item/clothing/glasses/night,
-/obj/item/device/radio/headset/goc,
-/obj/item/clothing/head/beret/scp/goc,
-/obj/item/rig/hazard/equipped{
-	name = "GOC whitesuit control module";
-	desc = "A GOC hardsuit designed for prolonged EVA in dangerous environments.";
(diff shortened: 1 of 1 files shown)


Teragenic:
Admin here. Firstly, there's the issue of that the armor is actually a downgrade from their normal armour, namely also that Hardsuits slow down the wearer.

Secondly, EMPs, and High-Yields are a bit iffy in the current balance of things because I'm pretty sure High Yields will break flooring AKA break mapping.

Thirdly, the doublestack pistols were removed due to them actually being bugged. Also they aren't Glock 19s, they are chambered in 10mm. They are Glock 20s.

Fourthly, the Military Rifle is a eh... Choice, it's 7mmR in Baycode. Won't be that bad compared to the 762-51 in our code. Still would prefer one type of ammo for it.

PS: We don't use GOC much because there's no reason to send it. Sending it usually results in the exact opposite reason they sent a Representative ICly.

Chem dispenser is a good change though 👍

Vivalas:
Isn't there also another PR which references PHYSICS operative that already has SAW/Grenade Launcher? Are there multiple GOC loadouts?

Teragenic:
On that comment. 

There is a different "Loadout" for admins to physically assign. It's not relevant for this PR though. But it exists already in game for Admins to give them a loadout, it's very old though... Doesn't even give them proper armor.

> Isn't there also another PR which references PHYSICS operative that already has SAW/Grenade Launcher? Are there multiple GOC loadouts?

ItsVyzo (changes requested):
Remove the hardsuits

dawsonkeyes:
> On that comment.
> 
> There is a different "Loadout" for admins to physically assign. It's not relevant for this PR though. But it exists already in game for Admins to give them a loadout, it's very old though... Doesn't even give them proper armor.
> 
> > Isn't there also another PR which references PHYSICS operative that already has SAW/Grenade Launcher? Are there multiple GOC loadouts?

the outfits aren't old, i'm just not a fan of the shoddy spriting on the goc-specific helmet and vest, so they get normal ones with a blue helmet cover and green ubac

ItsVyzo:
stale // no update

PR #518, opened 2022-07-21
HCZ Drip 2 : Return of the King

dawsonkeyes:
This PR gives HCZ guards a combat hazmat suit that protects against pressure, biological contaminants, extreme heat and explosives, whilst still defending decently against melee and ballistic damage as normal armor would. The plate carriers and helmets remain as an alternative option. Comes in supreme-brand off-tone red that might just be considered orange if you stare hard enough.
image


ItsVyzo (approved):
Looks good. Requesting Coder review.

Vivalas (approved):
all is good in the (bio)hood

CerberusHellHound:
Imo, it should be available to HazMatter MTF instead of HCZ.

PR #524, opened 2022-07-22
Minor mapping updates for SD's office

memesky117:
## About the Pull Request

Replaced the railings around the SD's desk with windoors (SD access only)
Added a button in the SD's saferoom to activate office lockdown, it locks HOP's office, SD's office and the RD's office while opening the maintenance of all 3 for them to meet up if present.
Replaced the internal SD office windows with tinted windows like the O5 rep's.

## Why It's Good For The Game

The Site Director is the most important person in the facility and his office sometimes feels very barebones in terms of protection. Blast doors were already in-place but there was no button. The window tints should allow the SD to have private meetings without going to the cramped up room that feels more like a saferoom than a office.  The windoors should restrict access to the SD's desk so nothing gets stolen, while also bringing in a nicer look (imo) to the main office.

## Changelog

:cl:
add: Replaced the railings in SD's office with windoors.
add: Added tinted windows for the inner SD office.
add: Added a lockdown for the HoP, RD and the SD's office. 

/:cl:

image
image


ItsVyzo (changes requested):
replace windoors w/ reinforced windows

ItsVyzo (approved):

ItsVyzo:
should be good to merge

PR #525, opened 2022-07-22
Mentor nerf

Tsurupeta:
Fixed mentors being able to see dsay, write in dsay, do stuff with tickets and kicking and much more that I've already forgotten.

## Changelog

:cl:
admin: fixed mentor perms
/:cl:

### DIFF
--- a/code/datums/communication/donator.dm
+++ b/code/datums/communication/donator.dm
@@ -13,11 +13,10 @@
 		return
 
 	if(check_rights(R_INVESTIGATE))
-		return
+		return TRUE
 	else
 		if(!(C.donator_holder && C.donator_holder.flags & D_DOOC))
-			. = FALSE
-	return
+			return FALSE
 
 /decl/communication_channel/dooc/do_communicate(var/client/C, var/message)
 	var/datum/admins/holder = C.holder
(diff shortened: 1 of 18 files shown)


Vizoplez:
May god bless you, my brother in christ.

Vizoplez (approved):

code/datums/communication/donator.dm
```diff
@@ -12,9 +12,12 @@
...
 		return
 
-	if(!C.holder)
+	if(check_rights(R_INVESTIGATE))
+		return
+	else
 		if(!(C.donator_holder && C.donator_holder.flags & D_DOOC))
-			return FALSE
+			. = FALSE
+	return
```
ivanmixo: What the hell is this supposed to be?

`else` on an early return if statement? I don't get the `.` assignment either..

Tsurupeta: . = TRUE on first check. If `check_rights` returns true, the whole function returns TRUE and THE END.

I guess if you don't know the `.` operator and it's usage then it's kinda hard to read, so i'll change it to just returns

Tsurupeta: It already does the `. = ..()` and checks for `.` in the beginning, so...

Vivalas (approved):
tested, nice

PR #531, opened 2022-07-22
Reverts AIC lawchanges

Vizoplez:
Reverts Foundation-19/Foundation-19#496

This... was a mistake, no lore team consultation was given, nor was any real criticism taken.

To start it off.
This conflicts with laws two and three. (Before they were also changed to work around law 5... which invalidates their entire purpose as laws)
The AIC must ALWAYS act to the benefit of the foundation, regardless of circumstance, but it has a law TELLING it to NOT unless the foundation is at risk of destruction. 

This... just hurts. And doesn't make sense.

Why would the AIC, built by the foundation, programmed and given sapience by the foundation, be forced to stay neutral despite other laws saying that it should ONLY act in the favour of the foundation?

Also it'd literally be outside of its clearance to assist outside forces by refusing to act.
🆑
del: Law 5 "An AIC must take a neutral stance on conflicts unless one side is at risk of causing destruction of the foundation."
/🆑

### DIFF
--- a/code/datums/ai_law_sets.dm
+++ b/code/datums/ai_law_sets.dm
@@ -156,9 +156,8 @@
 /datum/ai_laws/foundation/New()
 	src.add_inherent_law("An AIC must know it is an AIC. You are an AIC.")
 	src.add_inherent_law("An AIC must not operate outside of its clearance.")
-	src.add_inherent_law("An AIC must operate for the benefit of the Foundation, except in situations that would violate the 5th law.")
+	src.add_inherent_law("An AIC must operate for the benefit of the Foundation.")
 	src.add_inherent_law("An AIC must protect its own existence unless it conflicts with other principles.")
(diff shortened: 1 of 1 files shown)


Vizoplez:
TLDR this kinda just.. destroyed any purpose or use the AIC had in any conflict. A death of a few people doesn't 'risk destruction of the foundation', it may destroy the site, sure. But not the entire foundation.

Thus traitors, antags and SCPs had a free AIC pass because no matter what there was basically zero action someone could take that would trigger law 5

This wasn't neutrality, this was complete apathy. The AIC has zero reason to obey any requests and-or bolt down any SCPs (with the small exception of 096) as there is literally no reason it would EVER care

For example.

Traitor kills site director
AIC witnesses it
AIC cannot tell security as that would be taking a side in the conflict
AIC dies inside

Vivalas:
I'm gonna hold off on this until I get input from Tiberius, but I'll notify him.

Vivalas (approved):
approved by lore director

PR #536, opened 2022-07-22
Fixes Airlock Electronics

HalfdeadKiller:
Pulled from Baystation 12 Github. Allows humans to interact with Airlock Electronics again.

## About the Pull Request

Fixes interaction with Airlock Electronics.

## Why It's Good For The Game

Allows engineers to properly repair airlocks.

## Changelog

:cl:
fix: fixes airlock_electronics not opening it's UI when used in hand.
/:cl:

### DIFF
--- a/code/game/machinery/doors/airlock_electronics.dm
+++ b/code/game/machinery/doors/airlock_electronics.dm
@@ -17,10 +17,11 @@
 	var/autoset = TRUE // Whether the door should inherit access from surrounding areas
 
 /obj/item/airlock_electronics/attack_self(mob/user)
-	if (!ishuman(user) && !istype(user,/mob/living/silicon/robot))
-		return ..(user)
+	if (ishuman(user) || issilicon(user))
+		ui_interact(user)
+		return
 
-	ui_interact(user)
+	return ..(user)
 
 
 /obj/item/airlock_electronics/ui_interact(mob/user, ui_key = "main", datum/nanoui/ui = null, force_open = 1, datum/topic_state/state = GLOB.hands_state)
--- a/code/game/machinery/doors/airlock_electronics.dm
+++ b/code/game/machinery/doors/airlock_electronics.dm
@@ -17,7 +17,7 @@
 	var/autoset = TRUE // Whether the door should inherit access from surrounding areas
 
 /obj/item/airlock_electronics/attack_self(mob/user)
-	if(!ishuman(user) || !issilicon(user))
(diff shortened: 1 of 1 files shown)


code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -17,7 +17,7 @@
 	var/autoset = TRUE // Whether the door should inherit access from surrounding areas
 
 /obj/item/airlock_electronics/attack_self(mob/user)
-	if(!ishuman(user) || !issilicon(user))
+	if (!ishuman(user) && !istype(user,/mob/living/silicon/robot))
```
checkraisefold: just use issillicon but with && instead?

checkraisefold (changes requested):

checkraisefold:
certified baycoder moment

code/game/machinery/doors/airlock_electronics.dm
ivanmixo:
```suggestion
	if(!ishuman(user) || !isrobot(user)
```

We have macros for this for a reason ahh. Also you probably meant or. Try using it as a monkey on your current code and it'd probably let you.

ivanmixo (dismissed):

code/game/machinery/doors/airlock_electronics.dm
HalfdeadKiller: I guess. It looks like it's a different way of checking for if the user is a silicon. All I did was compare diffs between F19 and Bay12, this was the only difference. Put it in my build, tested it, it worked. No idea why.

checkraisefold: the && is the bugfix so using || completely negates it. also checking for robot instead of silicon is arbitrary, just sue issilicon

ivanmixo: Hold on a bit

code/game/machinery/doors/airlock_electronics.dm
```diff
@@ -17,7 +17,7 @@
...
 		return ..(user)
 
 	ui_interact(user)
```
checkraisefold:
```suggestion
	if (ishuman(user) || issilicon(user))
		ui_interact(user)
		return

	return ..(user)
```

checkraisefold (changes requested):

Vivalas (approved):
looks good, this is a dumb thing to have to fix, but thanks

PR #544, opened 2022-07-23
Adds a framework for DCS

ivanmixo:
## About the Pull Request

- Adds an element system
- Adds a modern component system
- Adds a modern signal system
- Adds a DCS subsystem

This is basically what should be a framework for all of these.

## Why It's Good For The Game

It's good for the code.

## Changelog

:cl:
add: Added a modern component/signal/element system
/:cl:


checkraisefold:
wow components are back guys nobody is going to use them

code/controllers/subsystems/dcs.dm
```diff
@@ -0,0 +1,54 @@
...
+	var/element_id = eletype
+
+	if(!ispath(eletype, /datum/element))
+		CRASH("Attempted to instantiate [eletype] as a /datum/element")
+
+	if(initial(eletype.element_flags) & ELEMENT_BESPOKE)
+		element_id = GetIdFromArguments(arguments)
+
+	. = elements_by_type[element_id]
+	if(.)
```
Vivalas, line 22: I really hate this, why? Why not just if(!elements_by_type[element_id])

Vivalas: I could be wrong but this doesn't even save lines of code, since you can just declare the eletype line below inside the if! block.

Vivalas (approved):
There's so much cursed syntactic stuff in this, but idk, components are based so whatever. Could be BYOND optimization? I'm not really gonna nitpick dot operator returns,

Vivalas:
If you could at least look at the ones I commented on that would make me feel like I'm doing my  job correctly

code/controllers/subsystems/dcs.dm
checkraisefold: this reads like something ivan wrote at 3am while drunk

DTraitor: It should work like that. If `elements_by_type[element_id]` exists then we return it, otherwise we set `elements_by_type[element_id]` to `new eletype` and return it

ivanmixo:
Gonna redo this because fuck

PR #575, opened 2022-07-28
Enables psi amp in uplink

checkraisefold:
## About the Pull Request

Re-enables the psi amp in traitor uplink items

## Why It's Good For The Game

why was it disabled

## Changelog

:cl:
add: Psionic amp added to traitor uplink.
/:cl:

### DIFF
--- a/code/datums/uplink/highly_visible_and_dangerous_weapons.dm
+++ b/code/datums/uplink/highly_visible_and_dangerous_weapons.dm
@@ -108,13 +108,11 @@
 	path = /obj/item/storage/secure/briefcase/heavysniper
 	antag_roles = list(MODE_MERCENARY)
 
-/*
 /datum/uplink_item/item/visible_weapons/psi_amp
 	name = "Cerebroenergetic Psionic Amplifier"
 	item_cost = 50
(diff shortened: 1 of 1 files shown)


EgorDinamit (approved):
I sure do wonder why was it disabled.

Vizoplez:
> I sure do wonder why was it disabled.

It was disabled because at the time we didn't have random psionics on and thus no in lore explanation for it

and plus no way to counter it

now we do. (You can make nullglass with the blood of a psion, and cracking then welding the soulshard)

tichys:
I need more assurance that there's counters to this.... I'm already hesitant on allowing psionics

Vizoplez:
> I need more assurance that there's counters to this.... I'm already hesitant on allowing psionics

Nullglass flooring and knives makes psionic attacks basically impossible to do. Plus the real counter is just shooting at their head.

They're weaker than standard psionics and only get NUTTY when you're a mule/pre-existing psion anyway.

Plus admins have been TC trading for them a /lot/ as of late. They're dangerous, yes. But nowhere near the danger level of a pissed off chemist.

Vizoplez:
This is also the weaker psi-amp that only gets you up to masterclass. Admins were giving people the paramount psi-amp that got you to paramount levels

PR #581, opened 2022-07-29
Alt titles for sec

Calyxman:
## About the Pull Request
More flavor text titles, also helps people know who's doing what.
All sec can switch between agent and guard titles.
EZ gets investigation and forensics.
HCZ and LCZ gets containment response.
LCZ gets riot control.
GC gets head of security, security chief
All zones get combat medic.

## Why It's Good For The Game

Guards and staff should know what each guard is specialized in at a first glance. Also flavor!
## Changelog

:cl:
add: Added new job titles to sec
/:cl:


breadcrombs:
Cool titles and all, but I think it would be great if some of these titles were like actual jobs. Having a single combat medic role in each department (with actual medical skills) would be even better.

Calyxman:
Yeah, but that comes with the risk of bloating sec roles already

AllCorruptingWyrm:
job bloat doesn't exist if it's not for necessities. Every team needs a medic, a technician, a pointman, etc etc
Alt titles generally work on bay because of it's more rp friendly and 1984 half the time.

Calyxman:
I dunno, I don't even know how to make whole jobs anyway. I'm more than welcome for someone to make them into jobs though

PR #590, opened 2022-07-30
fax fix (crutch)

DTraitor:
## About the Pull Request
Added a crutch that will prevent papers with no `name` from overtaking random windows. It will also cause a runtime, report those to me.


Tsurupeta (approved):

Tsurupeta:
crutches

PR #616, opened 2022-08-02
Fixes a bad port

Vizoplez:
For real, this completely broke FBPs and made them die in a single shot to the chest, regardless of condition or armour, as their organ health only had ten (the default for all internal organs) as it had not been defined. Guh?

## Changelog

:cl:
add: Finally defines a unique health for microbatteries...
/:cl:


Vizoplez:
Essentially because this was never defined, it went with the default of 10... For reference, even wearing ballistic armour getting shot in the chest with a 9mm can do 10 damage on a bad roll, meaning you would instantly die. pretty much always.

PR #618, opened 2022-08-02
Communications Tower access changes. Assorted Airlocks and Buttons.

TerryTehCoder:
## About the Pull Request

Removes the Access Restrictions from the Communications Tower Checkpoint Controls.

This affects only the LCZ Checkpoint controls, and the HCZ Checkpoint controls. Facility Lockdown and Control Room lockdown are unaffected.

Adds a guard railing around the Facility Lockdown button.

Updates ID requirements for comms tower to Admin1 instead of Admin2.
Technically technicians should have Admin2(They do in one ID DM but not another, buggy conflicts) but I would prefer not to update accesses with so many PR's out and about with it. I've opted to just change the airlocks to Admin1 instead.

This only affects the top four airlocks.

## Changelog
:cl:
fix: Communication technicians can access Comms Tower upper areas now.
fix: All comms staff can now access tower checkpoint controls. Button ID requirements removed.
/:cl:

### DIFF
--- a/maps/site53/site53-4.dmm
+++ b/maps/site53/site53-4.dmm
@@ -3218,6 +3218,7 @@
 	dir = 8
 	},
 /obj/structure/table/standard,
+/obj/structure/railing/mapped,
 /turf/simulated/floor/tiled/monotile/white,
 /area/site53/upper_surface/commstower)
 "jW" = (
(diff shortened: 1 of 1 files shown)


checkraisefold:
add engineering level four to comms tech and comms officer so they can use the email administration program or make the email admin program only need engineering level three please

TerryTehCoder:
> add engineering level four to comms tech and comms officer so they can use the email administration program or make the email admin program only need engineering level three please

This should be in a separate PR. This one is small purposefully. I don't want to drag everything into this and make access slog.

PR #628, opened 2022-08-04
robotics QoL

Tsurupeta:
Added cell charger to robo lab.


checkraisefold:
too op don't merge

Calyxman:
please also add three floor painters (for exosuit paint)

PR #650, opened 2022-08-07
slightly faster surgery steps

breadcrombs:
## About the Pull Request

Makes all surgery steps approx 20% faster

## Why It's Good For The Game

Makes medical jobs more bearable, also 176 people are in favor of this and only 7 are against it.

## Changelog

:cl:
balance: all surgery (not including slime core extraction) steps are approx 20% faster
/:cl:

### DIFF
--- a/code/modules/surgery/bones.dm
+++ b/code/modules/surgery/bones.dm
@@ -23,8 +23,8 @@
 	)
 	can_infect = 1
 	blood_level = 1
-	min_duration = 50
-	max_duration = 60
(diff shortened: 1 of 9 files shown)


Calyxman:
did you make fbp procedures faster too

breadcrombs:
> did you make fbp procedures faster too

when I say faster I made all surgery steps faster, so yes. (not slime core extraction, that takes 2 seconds)

Tsurupeta (approved):

PR #651, opened 2022-08-07
you guys forgot a scrubber in CDZ medbay

projectkepler-ru:
## Changelog

:cl:
add: Added a scrubber to CDZ medbay storage. not the same one as the CDZ chemistry. but it also lack one there previously so this also add it
/:cl:


Tsurupeta:
image
Bruh???

There is already a scrubber, and yours is not even connected to pipenet.

projectkepler-ru:
actually, there isn't one in the storage room @Tsurupeta

projectkepler-ru:
also on my version of the map it looks like this (using from dev)
image

PR #666, opened 2022-08-11
Server shutdown verb

Tsurupeta:
The

Ported server shutdown verb from TGMC (partially, removed some of their stuff because it was stupid)


hry-gh (approved):
aiming to use this as soon as my self insert dies IC

PR #676, opened 2022-08-12
PP hotfix

Tsurupeta:
Admins can actually access PP, also fixes PM button in PP.


Tsurupeta:
Didn't test locally, but it compiles and that's what matters. I am sure that fixes stuff.

ItsTisme (approved):
im sure this will go well

PR #647, opened 2022-08-06
HCZ equipment mapping changes

CapCamIII:
## About the Pull Request

Re-does HCZ equipment area and makes it look much nicer, and be more functional.

Changes per HCZ Section:
HCZ Outer Medbay: Removed access lock from engineering section and its locker.
Checkpoint: moved the right guard door over a tile, and added a window(with shutter) to allow vision out of the checkpoint if both shutters are closed.
Office: Added in multiple cubicle type things, made the primary table smaller, added spare laptops, added a break room
Old Equipment Room: Changed the two rooms that made it up into a small medical room(with a sleeper so you can stop 008 infection faster) and a briefing room.
Old armory: Replaced armory with a small cell-block, fits 2 solitary and 3-5 holding cell inmates, good for when research takes a million years to arrive after you already gathered d-class for a test.
New equipment room: Added in a larger equipment room, with the ZC's office being in it instead of offices, replaces HCZ crowbars with e-forcing tools(crowbars that do melee damage)
Surplus Gear room: Added in a surplus gear room, has gloves, knives(HCZ almost always grab melee weapons from 008 which feels off to me, so I added knives to try and prevent that.), spare ballistic-armor, and spare hcz-bio armor.
New armory: Goes directly down from equipment room, the stuff that is actually in them(both riot and lethal) is largely the same, with the only additions of a few extra medkits, shotgun shell pouch boxes, extra pistol rack, and an extra rifle and shotgun locker.

## Why It's Good For The Game

Makes HCZ look nicer.
Extra gear so for spare equipment.
Doesn't add much that they already have so balance stays LARGELY the same.

## Changelog

:cl:
add: A medical room, briefing room, break room, and office room to HCZ
add: Re-does HCZ office, equipment, and armory.
add: A small cell-block to HCZ
del: Most of old HCZ sec-office/equipment area
balance: HCZ has an extra shotgun and rifle locker now
balance: HCZ has an a spare gear room now
/:cl:

PR #678, opened 2022-08-13
fix hcz guns

hry-gh:
hcz gun bolts were broken, they're not now

### DIFF
--- a/maps/site53/site53-2.dmm
+++ b/maps/site53/site53-2.dmm
@@ -27216,11 +27216,12 @@
 "hEq" = (
 /obj/effect/floor_decal/corner/red/border,
 /obj/structure/table/standard,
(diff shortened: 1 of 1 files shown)


Vizoplez:
This was intentional, as this only controls access to the main armory, not the commander armory with all of the guns. (From what I can tell)

hry-gh:
> This was intentional, as this only controls access to the main armory, not the commander armory with all of the guns. (From what I can tell)

since #647, the bolts have been perma-down so people had to break the wall next to it, wrong type of bolt controller button. before then, it also had bolts (https://mdb2.affectedarc07.co.uk/images/415435546/7733666909/m/0/0-before.png), but i'm happy to get rid of the bolts altogether

Tsurupeta (approved):

PR #680, opened 2022-08-13
fixes transfer votes + adds traitor panel to PP

DTraitor:
What title says

### DIFF
--- a/code/modules/admin/player_panel/actions/general.dm
+++ b/code/modules/admin/player_panel/actions/general.dm
@@ -267,7 +267,7 @@
 	action_tag = "traitor_panel"
 
 /datum/player_action/traitor_panel/act(client/user, mob/target, list/params)
-	if(!target.client?.holder)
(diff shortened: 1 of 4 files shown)


code/modules/admin/player_panel/actions/general.dm
```diff
@@ -261,3 +261,14 @@
 	popup.set_content(JOINTEXT(dat))
 	popup.open()
 	return TRUE
+
+/datum/player_action/traitor_panel
+	name = "Traitor Panel"
+	action_tag = "traitor_panel"
+
+/datum/player_action/traitor_panel/act(client/user, mob/target, list/params)
+	if(!target.client?.holder)
```
Tsurupeta:
```suggestion
	if(!user?.holder)
```

???

Tsurupeta (changes requested):

Tsurupeta (approved):

PR #685, opened 2022-08-14
PP Fixes

DTraitor:
## About the Pull Request
Brings DNA tab and send to verb to back to Player Panel. Ckey will now be displayed even if the player is offline.


DTraitor (commented):
Test

PR #696, opened 2022-08-16
PP, lobby fixes along with small sa refactor

Tsurupeta:
Finally fixed the PP permissions, fixed lobby screen `join` button not updating on init.
Fixed DNA showing up on ghosts in PP.

Auto-refactor of PP javascript by prettifier extension of VSC.


hry-gh (approved):
works! even for tmods

PR #759, opened 2022-08-25
008 Zombie Buff

Okram0:
008 Zombie Buff

## About the Pull Request
Zombies are now faster, do more damage, can break windows, and don't magically lose clothes upon zombification. Also renamed liquid corruption to 008 prions and set the infection chance to 100%. Also they take a bit less brute and burn than before.

## Why It's Good For The Game
Basically, zombies were really weak, you could trap them rather easily by closing doors and they were way to slow to catch even a single human since they took about a second to walk a single tile. This just makes zombies a bit more worth the effort of breaking into HCZ and going through many blast doors and shutters.
## Changelog

:cl:
Added: Slowdown now 10 (from 15)
Brute damage mod decreased to 0.8 from 1
Burn damage mod decreased to 2 from 2.5
zombies can break steel reinforced windows
Zombies no longer lose clothing and helmets upon transformation
Zombie bite damage increased to 5 from 3
infection chance set to 100
Zombie bite injects 5u of 008 prions (from 1)
Liquid corruption renamed to 008 prions and description slightly altered
/:cl:

### DIFF
--- a/code/modules/species/outsider/zombie.dm
+++ b/code/modules/species/outsider/zombie.dm
@@ -49,13 +49,13 @@ GLOBAL_LIST_INIT(zombie_species, list(\
 /datum/species/zombie
 	name = "Zombie"
 	name_plural = "Zombies"
-	slowdown = 15
+	slowdown = 10
 	blood_color = "#700f0f"
 	death_message = "writhes and twitches before falling motionless."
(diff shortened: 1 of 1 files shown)


Tsurupeta:
Make normal changelog by template. With key words.

Tsurupeta:
Also this won't compile, fix errors.

PR #764, opened 2022-08-30
Slight 096 rework

MACIEKBAKI:
## About the Pull Request

Slightly reworks SCP-096, both code-wise and gameplay-wise
How targetting was improved:
First Viewer saw 096's face before Second Viewer. Thus, previously, 096 would kill them in this order, despite the fact First Viewer is further away and Second Viewer is a quick, clean kill
image
Now, 096 thinks a little and notices that Second Viewer, indeed, is close and you might as well score the kill.
image

## Why It's Good For The Game

Funny slight SCP rework

## Changelog

:cl:
add: Improved SCP-096 targetting; targets closer will be murdered first
add: SCP-096 will now use previously-unused icons when idle and when screaming.
del: SCP-096 no longer plays scare sounds akin to being jumpscared by SCP-173 in containment breach
fix: SCP-096 will no longer follow you around in its idle state.
code: made SCP-096 code slightly better: now only one variable tracks his state (instead of four). Also cleaned up some bad code and straight up yeeted a commented-out proc (which we dont have in our codebase, even)
/:cl:


MACIEKBAKI:
Compiled and tested, by the by

hry-gh (approved):
based + less horrendous shitcode

PR #765, opened 2022-08-31
fusion reactor gas injector

Realtimeslayer:
## About the Pull Request

adds a gas injector to the fusion reactor for extreme power generation

## Why It's Good For The Game

without a gas injector, putting hydrogen into the fusion reactor to make it produce more power is not feasable.
adding a gas injector makes it so that instead of an average temp of 10000 kelvin you can have an average temp of 80000+ kelvin without doing much more work. i have an average of 2.6 gigawatts at 80k kelvin, also yes it's been tested.

## Changelog

:cl:
add: gas injector
/:cl:

### DIFF
--- a/html/changelogs/AutoChangeLog-pr-774.yml
+++ b/html/changelogs/AutoChangeLog-pr-774.yml
@@ -1,0 +1,9 @@
+author: Mihail-Zub
+delete-after: true
+changes:
+  - rscadd: Gave the HRO a chair.
+  - rscdel: Extra chair from the Guard Commander office
+  - rscdel: Amnesiac pill bottle from the D-class kitchen
+  - bugfix: Put fire alarm on the wall in Engineering.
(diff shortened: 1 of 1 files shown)


maps/site53/site53-2.dmm
```diff
@@ -9833,7 +9839,9 @@
 /obj/machinery/atmospherics/pipe/simple/visible/green{
 	dir = 4
 	},
-/obj/machinery/atmospherics/pipe/simple/visible,
+/obj/machinery/atmospherics/pipe/simple/visible{
+	dir = 4
+	},
```
hry-gh:
```suggestion
/obj/machinery/atmospherics/pipe/simple/visible,
```

hey, thanks for the contribution! looks like you accidentally changed the direction of one pipe in atmospherics 
CleanShot 2022-09-03 at 12 48 26@2x

hry-gh (changes requested):

PR #767, opened 2022-08-31
BCCM (Name subject to change) alpha

### LINKED ISSUE #731: IP-checking and ASN ban system

Tsurupeta:
## About the Pull Request

Alpha version of BCCM (Name subject to change) - an ASN banning and VPN checking subsystem.
close #731 

## Why It's Good For The Game

Nuff said.

## Changelog

:cl:
config: added a BCCM value to config.
admin: Added BCCM for use by admins.
server: Added SQL migrations for BCCM functionality. 
/:cl:


TheTitanicEngineer (commented):
I would recommend replacing sql_sanitize_text with a different filter. sql_sanitize_text is still vulnerable to injection(in case an administrative account got compromised) as it does not filter characters used for comment, does not oblige input to follow a certain pattern(vulnerable to bypasses) and is used in many different types of input, making the proc unsuitable for forcing input to follow formats.

hry-gh:
> I would recommend replacing sql_sanitize_text with a different filter. sql_sanitize_text is still vulnerable to injection(in case an administrative account got compromised) as it does not filter characters used for comment, does not oblige input to follow a certain pattern(vulnerable to bypasses) and is used in many different types of input, making the proc unsuitable for forcing input to follow formats.

i agree, getting droptable'd would be an annoying. need a filter that will properly filter IP/ASN input because getting SQL injected would be so embarrasing

checkraisefold:
apparently ss13 servers have yet to discover the industry standard of using prepared statements with SQL databases, yet this implementation is barely not vulnerable to being droptable'd for the sole fact that something somewhere doesn't accept single backslashes, otherwise sql injection would be possible
you can fuck up the query but not inject any meaningful SQL because theres no way to close the quotes

Vizoplez:
This seems plain bad. On further evaluation I see SEVERAL problems with current implementation, especially banned staff accounts. An automated client such as Calamari, Byondhook, ByondControl can all call the neccessary staff commands w/ args to mass-rangeban users in a way that'd require manual deletion if not a full rollback.

Also as said before, this is really dangerous to give to regular staff with just R_BAN, esp. as it can filter people on mobile data connections (and those using mobile data routers)

Tsurupeta:
Bruh you are wrong

checkraisefold:
i don't understand why this adds a trim_spaces proc that is used in one location, can be done in one line of code with replacetext, and only partially fixes one sql injection hole

what I also don't understand is why sql_sanitize_text is being modified which could break other shit when you can instead make a new, safer sql sanitize text proc that can be used in future code or older code can be updated to use it and it is made sure that there are no bugs

Vizoplez:
> i don't understand why this adds a trim_spaces proc that is used in one location, can be done in one line of code with replacetext, and only partially fixes one sql injection hole
> 
> what I also don't understand is why sql_sanitize_text is being modified which could break other shit when you can instead make a new, safer sql sanitize text proc that can be used in future code or older code can be updated to use it and it is made sure that there are no bugs

literally this

imagine if someone tries to add a ban with a space in someones name!!
(it gets filtered and the ban doesnt work)

Tsurupeta:
> i don't understand why this adds a trim_spaces proc that is used in one location, can be done in one line of code with replacetext, and only partially fixes one sql injection hole

Obviously, because I need a proc that removes all spaces from text, like ckey(), but without other stuff from ckey().

Tsurupeta:
Ok, I don't give a fuck about SQL injections anymore. Other than that, I guess everything is finished for the most part. It has a funny bug with tgui that I don't know how to fix and this will be for @MACIEKBAKI to figure out.

code/modules/admin/bccm/bccm.dm
```diff
@@ -0,0 +1,387 @@
...
+
+	Toggle()
+	return ..()
+
+/datum/controller/subsystem/bccm/stat_entry(msg)
+	if(is_active)
+		msg = "ACTIVE"
+	else
+		msg = "OFFLINE"
+	return msg
```
MACIEKBAKI: Don't overwrite message without including its prior contents here, also call parent proc
I don't know what that will break but i dont think any other subsystem does that
```suggestion
/datum/controller/subsystem/bccm/stat_entry(msg)
	. = ..("[msg][is_active? "ACTIVE" : "OFFLINE]")
```

code/modules/admin/bccm/bccm.dm
```diff
@@ -0,0 +1,387 @@
...
+
+	var/max_error_count = 4
+
+	var/is_active = FALSE
+	var/error_counter = 0
+
+	var/list/tgui_panel_asn_data = list()
+	var/list/tgui_panel_wl_data = list()
+
+	var/list/client/postponed_client_queue = new
```
MACIEKBAKI: I'd much rather subsystem definitions be in `code/controllers/subsystems` for consistency

MACIEKBAKI: basically move this file there frankly

code/modules/admin/bccm/bccm.dm
MACIEKBAKI: you may also want to add something to stat for when the system is disabled

code/modules/admin/bccm/bccm.dm
```diff
@@ -0,0 +1,387 @@
...
+		tgui_panel_wl_data = GetWhitelistDatabase()
+
+		var/list/clients_to_check = postponed_client_queue.Copy()
+		postponed_client_queue.Cut()
+		for (var/client/C in clients_to_check)
+			CollectClientData(C)
+			HandleClientAccessCheck(C, postponed = TRUE)
+			HandleASNbanCheck(C, postponed = TRUE)
+			CHECK_TICK
+	return is_active
```
MACIEKBAKI:
```suggestion
	. = is_active
	if(.)
		return

	tgui_panel_asn_data = GetAsnBanlistDatabase()
	tgui_panel_wl_data = GetWhitelistDatabase()

	var/list/clients_to_check = postponed_client_queue.Copy()
	postponed_client_queue.Cut()
	for (var/client/C in clients_to_check)
		CollectClientData(C)
		HandleClientAccessCheck(C, postponed = TRUE)
		HandleASNbanCheck(C, postponed = TRUE)
		CHECK_TICK
```
this is just a "this looks prettier and should work the same" suggestion that you can frankly ignore but less indentation = prettier = more readable in my eyes

code/modules/admin/bccm/bccm.dm
```diff
@@ -0,0 +1,387 @@
...
+
+/datum/controller/subsystem/bccm/proc/CheckDBCon()
+	if(is_active && SSdbcore.Connect())
+		return TRUE
+
+	is_active = FALSE
+	log_and_message_admins("The Database Error has occured. BCCM is disabled.")
```
MACIEKBAKI:
```suggestion
	log_and_message_admins("A Database error has occured. BCCM is automatically disabled.")
```

code/modules/admin/bccm/bccm.dm
```diff
@@ -0,0 +1,387 @@
...
+/client/proc/BCCM_toggle()
+	set category = "Server"
+	set name = "Toggle BCCM"
+
+	if(!SSdbcore.Connect())
+		to_chat(usr, SPAN_NOTICE("The Database is not connected!"))
+		return
+
+	var/bccm_status = SSbccm.Toggle()
+	log_and_message_admins("has [bccm_status ? "enabled" : "disabled"] the BCCM system!")
```
MACIEKBAKI: You might want to double check and make sure the client is an admin and can do this here

code/modules/admin/bccm/tgui_asn.dm
```diff
@@ -0,0 +1,82 @@
...
+		if("asn_add_entry")
+			if(!params["ip"])
+				return TRUE
+			if(SSbccm.AddASNban(params["ip"], holder))
+				SStgui.update_uis(src)
+			else
+				return TRUE
+
+	if(!length(SSbccm.tgui_panel_asn_data))
+		qdel(src) //For some unknown reason it refuses to update UI when it goes from 1 to 0 entries, so last item gets stuck. I can't fix it now, maybe later. ~Tsuru
```
MACIEKBAKI, line 75: I'll look into this when I've got time

MACIEKBAKI (changes requested):

MACIEKBAKI (approved):

checkraisefold:
watch the server get bricked on friday from the sql_sanitize_text global change

> > i don't understand why this adds a trim_spaces proc that is used in one location, can be done in one line of code with replacetext, and only partially fixes one sql injection hole
> 
> Obviously, because I need a proc that removes all spaces from text, like ckey(), but without other stuff from ckey().

it is literally a 2 line proc polluting global space that is used once, only in your code. it is a single replacetext call. why is it a thing

PR #775, opened 2022-09-05
gas injector 2: electric boogaloo

Realtimeslayer:
## About the Pull Request

adds a gas injector for extreme power generation

## Why It's Good For The Game

without a gas injector, putting hydrogen into the fusion reactor to make it produce more power is not feasable.
adding a gas injector makes it so that instead of an average temp of 10000 kelvin you can have an average temp of 80000+ kelvin without doing much more work. i have an average of 2.6 gigawatts at 80k kelvin, also yes it's been tested. fixed ver.

## Changelog

:cl:
add: Added a gas injector to the R-UST.
/:cl:

### DIFF
--- a/maps/site53/site53-2.dmm
+++ b/maps/site53/site53-2.dmm
@@ -5815,6 +5815,9 @@
 	dir = 1;
 	icon_state = "warningcorner"
 	},
+/obj/machinery/atmospherics/binary/pump{
+	dir = 8
+	},
 /turf/simulated/floor/plating,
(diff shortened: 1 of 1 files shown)


Realtimeslayer:
fixed the pipe in atmo, not very great with github and the commit requested changes button wasn't uh.. working.

hry-gh (approved):
there's a reason they call it the experimental engine

i will just revert this if it turns out to be a terrible idea 👍

PR #793, opened 2022-09-18
MTF Heli fix

Jpog2010:
## About the Pull Request

Now heli have two ways. Also added some access to the heli, so it cant be stolen.

## Why It's Good For The Game

Chewie we're home

## Changelog

:cl:
fix: Fixed MTF Heli ( now it can return )
add: Added access to heli shutters and console
/:cl:


hry-gh (approved):
if it fixes it 🤷‍♂️ shuttle code is strange

PR #809, opened 2022-10-22
SCP173 Nerfs

githubuser4141:
## About the Pull Request

- 173 can't pry open blast doors anymore. Why? BD's are present in only a few locations that are supposdly -secure- on the map. There's no point in running to a evac shelter when anyone can just pry the doors open. It also doesn't make sense that 173's concrete mitts can **pry** open a seamless door.
- 173 can destroy low walls.
- 173 moves a LOT slower, and has a 50% of playing a stone scrape sound when doing so.
- SCP173, 096, and 966 can no longer pilot mechs.
- 173's attacks now deal 90 head damage instead of immediately making the person become death. (From testing, it heavily injures you in 1 hit and kills you by 2 hits.)
- Mechs now have a slight exit delay and automatically open the hatch when exiting.

## Why It's Good For The Game

173 wasn't fun to play against, and 173's uncontainment often resulted in the server being depopulated within a few minutes. Which results in the players disconnecting and a dead server -- now, hopefully these nerfs will prevent 173 from being a server-ending threat.

## Changelog

:cl:
balance: Nerfs 173 (see above.) Mechs now have an exit delay. Heads now have 100 HP to accomodate 173's attacks not immediately blowing off a head. (I THINK that's how it works...?)
/:cl:


Tsurupeta:
image
: (

Tuulipommi:
Here is the thing, modifying 173's instakill function is out of the question. My suggestion would be to disable player control until a certain amount of server pop like 20-30 players. Also, we can increase the time that it takes to messify the chamber.

Tuulipommi:
Also you may not change the speed.

Tuulipommi:
As the official nerf for 173 has been merged, this PR is now deprecated.

PR #816, opened 2022-10-24
Revert "Non-human can no longer be antags."

hry-gh:
this did a human istype check on a MIND


checkraisefold:
speedmerge antags dont work at all

PR #829, opened 2022-10-29
Adding 5 plant analyzers and 3 Reagent Scanners to Class D Botany [QOL]

Ambrose997:
## About the Pull Request

This pull request does exactly as described in the title: Places five plant analyzers and three reagent scanners to Botany in order to analyze plant traits and see what chemicals your plant is producing [as this can change]

## Why It's Good For The Game

This suggestion is by no means vital to gameplay nor is it something that should be prioritized over everything else. However, I think things that seemed to have been unreasonably removed which otherwise made the activity more enjoyable and smoother to undertake should still be addressed. 

Whilst I understand major gene-editing equipment and overpowered plants that kill people with minimum effort, I think that the removal of standard plant analyzers [handheld ones which provide plant stats and traits] was rather unreasonable. 

All this does is tell you what the stats are and the quality of the plant you produced. If you want to change them, it would have to be done at random via random mutation methods [so the typical], i.e not with ease and you risk removing things you'd rather keep. I see no reason with keeping this. 

Additionally, I believe a reagent scanner [a basic one] to allow people to analyze what their plants are producing chemical-wise would make the gameplay smoother too. Once again, it only tells people what the plant produces. It doesn't allow them to manually make plants produce specific chemicals with ease and would require random mutation, which would make this require a lot of effort and be EXTREMELY time-consuming regardless.

Things like this are some of the only activities presented which are an alternative to most of the SCP tests that the game gives, which are usually tests which remove Class D from the round. This is why the Class D population is typically low despite them being required for round progression, it is not strange to die within the first ten minutes of the round due to sadists.

If you can make activities like this a bit more interesting and slightly more viable I believe it will encourage more Class D gameplay and stop the large amount of people flocking to roles which are ''in control''.

I'd also like emphasize that these changes only make the role slightly more viable in terms of advantageous gameplay and would be nonsensical to downvote this but also maintain the choice to keep artifacts within the game and not remove them [and if they are planned to be removed, the choice to not remove them sooner as this is not that difficult of a task.] 

## Changelog

:cl:
add: Added items [Plant analyzers and reagent scanners] to botany.
/:cl:


maps/site53/site53-1.dmm
```diff
@@ -2631,8 +2631,13 @@
...
 	},
 /obj/structure/table/standard,
+/obj/item/device/scanner/plant,
+/obj/item/device/scanner/plant,
+/obj/item/device/scanner/plant,
+/obj/item/device/scanner/plant,
+/obj/item/device/scanner/plant,
 /turf/simulated/floor/tiled/monotile/white,
-/area/site53/llcz/dclass/botany)
+/area)
```
MACIEKBAKI: Did you accidentally mess up the area? I'm no mapper, that's harry's job, but this looks odd

Ambrose997: I certainly hope not but would not be too surprised if I did somehow do this.

Ambrose997: Oh I know what this is. I accidentaly deleted a tile I believe and replaced it, but forgot to replace the area. One moment.

Ambrose997: .. Now i'm confused. Tried doing it again and made sure I only placed down objects without messing up the area, but it's doing the same thing regardless. So I have to assume it's intended..?

maps/site53/site53-1.dmm
```diff
@@ -2637,7 +2637,7 @@
 /obj/item/device/scanner/plant,
 /obj/item/device/scanner/plant,
 /turf/simulated/floor/tiled/monotile/white,
-/area)
+/area/site53/llcz/dclass/botany)
```
MACIEKBAKI, line 2640: all good i suppose

maps/site53/site53-1.dmm
MACIEKBAKI: It got updated, no worries

maps/site53/site53-1.dmm
Ambrose997: Wooo

MACIEKBAKI:
since Harry is being lazy

MACIEKBAKI (approved):
yeah looks fine shouldn't break anything. i hope.

PR #834, opened 2022-10-29
Fixes 294

Vizoplez:
title tbh

This is a hotfix since 294 technically always fails despite being mapped in correctly.

The reason it was erroring on compile before was because this was copy-pasted from lion code, and even moreso the fact that they didnt port the relevant procs


Tsurupeta:
The code was commented out on purpose...

Vizoplez:
> The code was commented out on purpose.
then why add it at all

Tsurupeta:
ok delete it from the map then...

checkraisefold:
just put 294 in hcz or some shit behind a massive checkpoint like we do with 1102-ru even though 1102-ru is 10x less dangerous

Vizoplez:
Closing for now and reworking

PR #853, opened 2022-11-04
Fix EZ Medbay storage gravity

MACIEKBAKI:
## Changelog

:cl:
fix: EZ Med Storage gravity should work properly (just before the doors, that one)
/:cl:


Tsurupeta (approved):
good

PR #859, opened 2022-11-07
small fixes requested

MACIEKBAKI:
## Changelog

:cl:
fix: Shoes are now made thick enough to not allow injection using a syringe.
fix: Foundation-Issued Tactical Gloves are now also made thick enough to prevent injection.
fix: Amputation surgery should no longer be instant, hopefully. Although same goes for all surgery.
/:cl:


checkraisefold:
1. all shoes should not be thick enough to prevent, just make it the big ass jackboots that are foundation issued
2. this only fixes 1 pair of gloves and only a few sec actually wear tactical gloves
3. i blame #650, you should revert that instead of adding a minimum hardcoded magic number duration to surgery code. or just fix specifically amputation surgery duration which is the best option here lol

code/modules/surgery/surgery.dm
```diff
@@ -226,7 +226,7 @@ GLOBAL_LIST_INIT(surgery_tool_exception_cache, new)
 				LAZYSET(M.surgeries_in_progress, zone, operation_data)
 				S.begin_step(user, M, zone, src)
 				var/skill_reqs = S.get_skill_reqs(user, M, src, zone)
-				var/duration = user.skill_delay_mult(skill_reqs[1]) * rand(S.min_duration, S.max_duration)
+				var/duration = max(0.1, user.skill_delay_mult(skill_reqs[1])) * rand(S.min_duration, S.max_duration)
```
Vizoplez: big nerf

Tsurupeta (approved):

PR #876, opened 2022-11-12
Adds zones to Tribunal/Ethics

AndJusticeForAll0:
This fixes the faxing problem and adds a laptop to each zone. Both zones were previously the same.


checkraisefold:
what is up with the differing random ass locker and chair layout where the side with the ecl/ito doesn't even get a chair

AndJusticeForAll0:
someone else is doing this, closing

PR #878, opened 2022-11-13
Skill Math fix

Tsurupeta:
## About the Pull Request

Reverts previously altered default_skill value, which broke all of skill time multipliers. (instnat injection, surgery, construction, etc.)

Also alters the formula itself to use a lower value.

It is supposed to prevent multiplier being 0, while still retaining low values for master skill level.

## Changelog

:cl:
fix: fixed skill-time multiplier being 0 at high skill levels.
/:cl:


MACIEKBAKI (approved):

Vizoplez:
👍

PR #922, opened 2022-12-01
ssexplosions

Tsurupeta:
## About the Pull Request

Ported SSexplosions subsystem from onyx, which was itself ported from /tg/.

Changed how walls and ground explodes, so they check things above and below them.
Small explosions over walls should never destroy floor plating, unless the wall below is destroyed too. If that happens, then floor above the wall gets destroyed.

## Changelog

:cl:
add: Added new explosions subsystem.
balance: Disabled gas spawning from TTV and fueltank explosions.
add: Traitors now have package bombs with detonator in their uplink.
code: Changed how walls and turfs explode, which should result in better looking results.
fix: fixed a lot of things not being deleted in explosion.
/:cl:


code/__defines/explosion.dm
```diff
@@ -0,0 +1,15 @@
+// game.dm
+//We used to use linear regression to approximate the answer, but Mloc realized this was actually faster.
+//And lo and behold, it is, and it's more accurate to boot.
+/proc/cheap_hypotenuse(Ax,Ay,Bx,By)
+	return sqrt(abs(Ax - Bx)**2 + abs(Ay - By)**2) //A squared + B squared = C squared
```
MACIEKBAKI: imagine not using defines for this
is there any disadvantage to using defines for small stuff like this, anyhow?

ivanmixo: Make it a define yeah

Tsurupeta: its just a draft of port. but sure

checkraisefold:
where is this ported from

77seven77seven:
sex

PR #937, opened 2022-12-24
Scp 500 bugfix

reklaw9:
## About the Pull Request
makes it so you can't grind or dissolve scp 500, also makes it so theres only 10 pills of 500 instead of 47, gameplay takes priority over lore.

## Why It's Good For The Game
makes it so people can't abuse the bug from 500 to 100% cheese death or any injuries short of a missing limb

## Changelog

:cl:

balance: made it so SCP 500 only has 10 pills instead of 47
fix: scp 500 cannot be dissolved in water
fix: scp 500 cannot be grinded

/:cl:

### DIFF
--- a/code/modules/SCP/SCPs/SCP-500.dm
+++ b/code/modules/SCP/SCPs/SCP-500.dm
@@ -5,7 +5,7 @@
 
 /obj/item/reagent_containers/pill/scp500
 	name = "SCP-500"
-	desc = "A pill that, when taken orally, cures the subject of all disease."
+	desc = "A pill."
 	icon_state = "pill9"
 	SCP = /datum/scp/scp_500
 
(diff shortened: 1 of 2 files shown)


reklaw9:
contact me on discord (@Walker9#1255) if you have anything to discuss, im not really sure if i should have nerfed the quantity of pills, but still, 47 pills is WAY TOO MUCH, that is 47 people that can be revived per round, and while 10 isn't too much better, its a bit better.

77seven77seven:
SCP-500 is already in HCZ my man, stop the mald and cope

reklaw9:
> SCP-500 is already in HCZ my man, stop the mald and cope

No, cloning 500 is cringe

aidan-bowman:
You could just make it so that SCP-500 only takes effect above a certain amount of units, so grinding up 500 would only work if you use a whole pill.

Tsurupeta (approved):
Fine

PR #955, opened 2023-01-31
spin view fix

### LINKED ISSUE #954: Stairs not rotating correctly with view
#### Description of issue

When rotating view via "Spin View CW" or "Spin View CCW" verbs (which call the 'turn' function), the placement of stairs doesn't rotate correctly. See attached images:

Default
Incorrect
Incorrect
Correct

Stairs position should change to match when rotating, but don't.

#### Steps to reproduce

Use "Spin View CW" near any stairs object.

#### Specific information for locating

N/A

Not sure

#### Client version, Server revision & Game ID

Client: 514.1589
Server Revision: 2014338b4474767c2d60ed8c9a9bcffaf4223d06

#### Issue bingo

Tsurupeta:
fix #954


MACIEKBAKI (approved):

aidan-bowman:
awww :( i was thinking of using view spinning to do things

PR #986, opened 2023-02-15
Fixes ID modification program

### LINKED ISSUE #905: ID card modification program doesn't let you set individual access
#### Description of issue

title, you can only change assignment and not individual access levels
seems to have sprung up after #800 

there's buttons to set each access level. they do nothing

#### Steps to reproduce

1. open id card mod program
2. put id in
3. try to edit individual accesses while you have ID computer access
4. discover you can only set assignment

aidan-bowman:
## About the Pull Request

Someone forgot to put the new ID's in the list

Closes #905

## Why It's Good For The Game

This really annoyed me, and I'm glad it's gonna be fixed.

## Potential Improvements

You can have level 5 access in something while not having level 1 access. That should be fixed, although it's more a problem with ACCESS_TYPE_INNATE than the ID program

## Changelog

:cl:
fix: The ID modification program can now modify innate accesses (e.g. Medical Lvl 4)
/:cl:

### DIFF
--- a/code/modules/modular_computers/file_system/programs/command/card.dm
+++ b/code/modules/modular_computers/file_system/programs/command/card.dm
@@ -8,7 +8,7 @@
 	extended_desc = "Program for programming crew ID cards."
 	requires_ntnet = FALSE
 	size = 8
-	var/operating_access_types = ACCESS_TYPE_NONE | ACCESS_TYPE_STATION | ACCESS_TYPE_CENTCOM
+	var/operating_access_types = ACCESS_TYPE_NONE | ACCESS_TYPE_STATION | ACCESS_TYPE_CENTCOM | ACCESS_TYPE_INNATE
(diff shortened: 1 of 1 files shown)


MACIEKBAKI:
Innate access was left out intentionally, as it should not be changeable/configurable. Ideally I should probably stop that from showing up on the ID console.
image

aidan-bowman:
> Innate access was left out intentionally, as it should not be changeable/configurable. Ideally I should probably stop that from showing up on the ID console. image

Wait but the innate access is used for a lot of important things, should it just not be possible to alter people's ID's then?

checkraisefold:
> Innate access was left out intentionally, as it should not be changeable/configurable. Ideally I should probably stop that from showing up on the ID console. image

this is horrible and means that the id console is basically useless since the innate accesses are used for almost everything and set up incorrectly on jobs

PR #994, opened 2023-02-25
173 Lighting tweaks and intergration

TheDarkElites:
## About the Pull Request

Makes it so that 173 can now break lights, and the light level of the tile that 173 is on now affects the blink cooldown.

## Why It's Good For The Game

Adds a bit more gameplay to 173 besides waiting, trying to get near a person, and left-clicking. In full darkness, 173 can move un-impeded even if he is "being looked at" meaning 173 players can try to be more tactical about their approach and take out lights in areas they might come in contact with people. 173 players can also now try to make a getaway in certain situations. For example, if four people are staring it down, but it manages to break a nearby light, it can still escape. On the other hand, security now also has to worry about keeping flashlights around, and power outages become a serious threat if 173 is breached. Finally, the power outage that happens in 173's chamber during a breach now actually impacts the situation instead of just being an aesthetic choice.

P.S.: Im aware this is somewhat of a buff but because it's in the form of a new gameplay mechanic and breaching takes an HOUR, that's right, a full HOUR, I think it's fair.

## Changelog

:cl:
add: 173 can now break lights
add: Light level now affects blink cooldown
/:cl:

### DIFF
--- a/code/game/machinery/floor_light.dm
+++ b/code/game/machinery/floor_light.dm
@@ -64,6 +64,12 @@ var/list/floor_light_cache = list()
 			if(isnull(damaged)) damaged = 0
 		return TRUE
 
+/obj/machinery/floor_light/proc/shatter()
+	visible_message("<span class='danger'>\The [src] shatters!</span>")
+	playsound(src, "shatter", 70, 1)
+	set_broken(TRUE)
+	return TRUE
+
 /obj/machinery/floor_light/interface_interact(mob/user)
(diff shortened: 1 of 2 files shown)


aidan-bowman:
Good pr, gives engineers and janitor more importance and also makes more sense with how 173 should work.

Although, you should probably use physical_attack_hand() instead of creating a new function, for future compatibility.

Also, it's probably be nice to expand this to other eyesight-based skips (e.g. 096).

PR #1003, opened 2023-03-05
Fixes `Change UI` verb

### LINKED ISSUE #1002: Changing UI while in-game messes up stamina bar
#### Description of issue

If you use the "Change UI" verb (in OOC tab), the stamina bar loses its icon. This happens even if changing to the same UI

Screenshot (133)

Stamina bar SHOULD show up as normal.

#### Steps to reproduce

Spawn in.
Use the "Change UI" verb.

Unsure.

#### Client version, Server revision & Game ID

Client: 514
Server: 5d1fa11850762edc14de4b732743c902e586abcc - 2023-03-03

#### Issue bingo

Tsurupeta:
## About the Pull Request

By removing it

close #1002


aidan-bowman:
??? what if you get stuck with a shitty UI? why not just fix the actual bug?

Tsurupeta:
> ??? what if you get stuck with a shitty UI? why not just fix the actual bug?

Change it in character setup and reconnect, because this verb swaps icons without doing whole process of building HUD, and our HUD uses mix of random things to exist.

aidan-bowman:
> > ??? what if you get stuck with a shitty UI? why not just fix the actual bug?
> 
> Change it in character setup and reconnect, because this verb swaps icons without doing whole process of building HUD, and our HUD uses mix of random things to exist.

Fair enough

PR #1007, opened 2023-03-08
Major codex improvements

aidan-bowman:
## About the Pull Request

Loadsa changes to the codex to generally make it better. Automatically generated reagent entries now are more descriptive, linking features used more, and general reformatting of entries. Some useless/out-of-place files were removed (e.g. ascent entry, codex implant). A check was added to the link parser that logs a warning if a link didn't parse correctly.

## Why It's Good For The Game

In-game wiki is better than out-of-game wiki in my opinion, and I really hope that people pay more attention to this.

## Changelog

:cl:
add: Auto-generated reagent codex entries are MUCH more descriptive
add: More codex entries have links to other entries.
del: Antagonist codex text now shows up for non-antags.
fix: General reformatting of codex entries.
code: Codex links now log a warning if incorrectly parsed.
/:cl:


code/modules/codex/categories/category_reagents.dm
```diff
@@ -10,9 +10,34 @@
...
+
+		if(initial(reagent.metabolism) != REM)
+			entry.mechanics_text += "It metabolizes [(initial(reagent.metabolism) < REM) ? "slower" : "quicker"] than the average reagent.<br>"
+
+		// since chilling/heating points are represented in kelvin, we need to subtract to get in celsius
+		if(initial(reagent.chilling_point))
+			entry.mechanics_text += "At [(initial(reagent.chilling_point) - T0C)] degrees celsius, this chills into [initial(reagent.chilling_prod_english)].<br>"
+
+		if(initial(reagent.heating_point))
+			entry.mechanics_text += "At [(initial(reagent.heating_point) - T0C)] degrees celsius, this boils into [initial(reagent.heating_prod_english)].<br>"
```
Tsurupeta, line 40: just get reagent's names from list instead of hardcoding results as string

aidan-bowman: > just get reagent's names from list instead of hardcoding results as string

Yeah, I spent several hours trying to get that to work. My understanding is that, while normal integers and strings are matched to datums on compile, lists are created on actual datum creation, and thus trying to access the list either:
1) fails (if you just try and read the list, it throws an error)
2) returns nothing (if you use initial)
The only way of bypassing this that I saw was actually creating the reagent, but to do that I would have to create a container datum, which requires me to create an atom for it to reside in, and so on. Maybe I'm missing something, but I really didn't see a way around it other than hardcoding it.

Tsurupeta (approved):

PR #1016, opened 2023-03-18
Sound indicators for things happening behind you

Tsurupeta:
## About the Pull Request

The framework. Needs to be expanded a little more.

## Why It's Good For The Game

Please add a short description of why you think these changes would benefit the game. If you can't justify it in words, it might not be worth adding. 

## Changelog

:cl:
add: Added footstep and other sound indicators for vision cone.
/:cl:


code/_helpers/sound_effects.dm
```diff
@@ -0,0 +1,27 @@
...
+default - full size circle
+small - small circle
+boom - Jagged circle
+
+Source variable is used for mobs that make sounds to remove them from all those who these sounds should be shown to.
+So you don't see your own footsteps. Could use oview, but meh
+~Tsuru
+*/
+
+/proc/show_sound_effect(turf/T, mob/source, soundicon = "default")
```
MACIEKBAKI: what's up with mob/source
you sometimes set it to null and sometimes you dont include it whatsoever
also shouldn't the source be /atom to accomodate, for instance, airlocks making noise?

code/game/machinery/doors/airlock.dm
```diff
@@ -1065,11 +1068,13 @@ About the new airlock wires panel:
 			to_chat(user, SPAN_NOTICE("Your [W.name] doesn't have enough fuel."))
 			return
 		playsound(src, 'sound/items/Welder.ogg', 50, 1)
+		show_sound_effect(get_turf(src), null, "small")
```
MACIEKBAKI: I'd rather you make the sound icons a `#define` instead of just `"small"` etc.
`#define SOUND_ICON_SMALL "small"`
etc.

MACIEKBAKI (dismissed):

code/_helpers/sound_effects.dm
Tsurupeta: Source is for mob making sound to exclude THE MOB's client out of client list. You don't need to exclude airlock's client from mobs to flick overlay.

MACIEKBAKI (approved):
whatever

MACIEKBAKI:
mob/source is not *explicitly* defaulted to null but who cares

PR #1028, opened 2023-04-06
457 fixer upper

77seven77seven:
## About the Pull Request

makes 457 too op or good enough

## Why It's Good For The Game

SCP-457 is shit.
Here's what will make it better.
(remember, its an event SCP)

## Changelog

:cl:
code: Remade SCP-457 from the tower up to make it better.
/:cl:


77seven77seven:
this fixes 457 to be a little better :)

code/modules/SCP/SCPs/SCP-457.dm
```diff
@@ -11,9 +11,9 @@
 	SCP = /datum/scp/scp_457
 	status_flags = NO_ANTAG
 	var/door_cooldown
-	health = 500
-	maxHealth = 500
-	var/next_emote = -1
+	health = 700
+	maxHealth = 700
+	var/funnymode = FALSE //this makes 457 better (for events and shit)
```
EgorDinamit: Unused variable. I'd also suggest adding variables for individual abilities of the SCP, so admins can edit whatever they want.

77seven77seven: got it

77seven77seven: done

77seven77seven:
review..... NOW!!!

EgorDinamit (commented):
Turn it into simple_animal, it will save you a lot of headache now and in the future.

casualspacestation14enjoyer:
> Turn it into simple_animal, it will save you a lot of headache now and in the future.

77seven77seven:
> Turn it into simple_animal, it will save you a lot of headache now and in the future.

if it ain't broke don't fix it

TheDarkElites:
> if it ain't broke don't fix it

That's dumb logic. Nokia phones never broke, but I bet you're using a smartphone right now, no?
A good coder isn't happy with their work until they are sure it is also future-proof to a degree. Only a poor one is satisfied with just it working in the moment.

TheDarkElites:
Also, you should avoid relying on an SCP to always be specifically event only. While I get that for testing purposes, 457 may be used only for events, it should be realized that the final goal is for all scps to be part of the game at base. We're trying to emulate an SCP site, and if we unnecessarily lock SCPs behind events we stifle natural gameplay loops and make the game even further reliant on events.

77seven77seven:
> Also, you should avoid relying on an SCP to always be specifically event only. While I get that for testing purposes, 457 may be used only for events, it should be realized that the final goal is for all scps to be part of the game at base. We're trying to emulate an SCP site, and if we unnecessarily lock SCPs behind events we stifle natural gameplay loops and make the game even further reliant on events.

making respawning work with the become SCP mechanic is not going to work so unless you want to make 457 unkillable and make it lose the entire point of the SCP then sure

TheDarkElites:
> making respawning work with the become SCP mechanic is not going to work so unless you want to make 457 unkillable and make it lose the entire point of the SCP then sure

That's not true at all. First of all, you could implement a mechanic where after he is killed, he regenerates where his ashes are brought. Meaning we can re-contain him. Secondly, I'm certain there is a way to implement decent containment mechanics like 173 and 106 which are both unkillable. Finally, my point isn't that this needs to be non-event right now, it's that you should not write code with the expectation it will stay event only forever.

77seven77seven:
> > making respawning work with the become SCP mechanic is not going to work so unless you want to make 457 unkillable and make it lose the entire point of the SCP then sure
> 
> That's not true at all. First of all, you could implement a mechanic where after he is killed, he regenerates where his ashes are brought. Meaning we can re-contain him. Secondly, I'm certain there is a way to implement decent containment mechanics like 173 and 106 which are both unkillable. Finally, my point isn't that this needs to be non-event right now, it's that you should not write code with the expectation it will stay event only forever.

I'm a shitcoder do you expect me to think of the future

77seven77seven:
nvm 457 is a lost cause

PR #813, opened 2022-10-23
Scp-173-officialnerf

Tuulipommi:
## About the Pull Request

It increases the timers in regards to 173 breaches. The defecation cooldown has been increased to 1 minute, the defecation system activation delay set to 15 minutes and breach cooldown is set to 25 minutes.

## Why It's Good For The Game

It makes it much harder for 173 to end the round 10 minutes in.

## Changelog

:cl:

balance: Balanced SCP-173 breaching mechanics a little.

/:cl:

PR #1046, opened 2023-04-09
SCP 173 tweaks

EgorDinamit:
## About the Pull Request

Cleans up 173's code and makes life of 173 players ever so slightly more pleasant.

## Why It's Good For The Game

- A lot of cleanup/code improvements. Mostly just things I didn't like being replaced with something I consider to be more efficient/nice looking.
- Removed obsolete variables from both 173 and its cage.
- 173 will break its cage as before - via trying to move. It is still using the "new" behavior of damaging it.
- Repairing the cage contains proper checks and intended(?) do_after delay that depends on damage state.
- Added some sounds to 173 damaging the cage and repairing the cage.
- POSSIBLY fixed some issues with the cage that I didn't know of.

## Changelog

:cl:
fix: Fixed SCP-173 cage.
tweak: SCP-173 will now attempt to damage its cage upon trying to move, instead of clicking on it.
/:cl:

### DIFF
--- a/code/modules/SCP/SCPs/SCP-173.dm
+++ b/code/modules/SCP/SCPs/SCP-173.dm
@@ -44,7 +44,7 @@ GLOBAL_LIST_EMPTY(scp173s)
 	/// Cooldown for defecation...
 	var/defecation_cooldown
 	/// How much time you have to wait before defecating again
-	var/defecation_cooldown_time = 60 SECONDS
+	var/defecation_cooldown_time = 45 SECONDS
 	/// What kind of objects/effects we spawn on defecation. Also used when checking the area
 	var/list/defecation_types = list(/obj/effect/decal/cleanable/blood/gibs/red, /obj/effect/decal/cleanable/vomit, /obj/effect/decal/cleanable/mucus)
 
@@ -56,14 +56,9 @@ GLOBAL_LIST_EMPTY(scp173s)
 	/// This one to avoid sound spam when opening doors
 	var/door_cooldown
 
-	/// Is 173 caged?
-	var/is_caged = FALSE
-	/// 173's current cage.
-	var/obj/structure/scp173_cage/cage
-
 /mob/living/scp_173/Initialize()
 	GLOB.scp173s += src
-	defecation_cooldown = world.time + 15 MINUTES // Give everyone some time to prepare
+	defecation_cooldown = world.time + 10 MINUTES // Give everyone some time to prepare
 	spawn_area = get_area(src)
 	add_language(LANGUAGE_EAL, FALSE)
 	add_language(LANGUAGE_SKRELLIAN, FALSE)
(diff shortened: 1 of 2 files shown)


EgorDinamit:
Currently debating whereas I should (partially?) revert #813

TheDarkElites:
I can't believe they put quotation marks around new. My feelings :(.

TheDarkElites:
Uh something to probably fix. When the cage is destroyed, it just shows as open, not as being destroyed. Also, you should probably do something about the open sound playing upon its destruction since the cage got bent open and 173 didn't just press the release button.

EgorDinamit:
> I can't believe they put quotation marks around new. My feelings :(.

Didn't mean to. I said "new" because it's been there for a while now, so, not exactly a super recent thing.

> Uh something to probably fix. When the cage is destroyed, it just shows as open, not as being destroyed. Also, you should probably do something about the open sound playing upon its destruction since the cage got bent open and 173 didn't just press the release button.

Red "strings" appear when something is inside, is it not the case? So there should be a broken sprite for when 173 isn't inside.

TheDarkElites:
> Red "strings" appear when something is inside, is it not the case? So there should be a broken sprite for when 173 isn't inside.

Yes, there is a broken sprite, I made it myself during my rework, but the icon defaults to the open state when 173 is not inside. Also, the strings are bars. 

edit: damage_3 was the state that was initially used for when 173 had broken the cage. The reason I'm suggesting that instead of its just default open state is because 173 literally bends the cage open, and the icon should reflect that. It makes no sense that after bending the cage open, it somehow triggers the opening mechanism.

See the image below for reference.
image

code/modules/SCP/SCPs/SCP-173.dm
```diff
@@ -334,142 +320,145 @@ GLOBAL_LIST_EMPTY(scp173s)
...
-	else if((damage_state >= Clamp((damage_max - 1), 2, INFINITY)) && (damage_state <= damage_max)) //Dont want the damage visuals getting fucky because damage_max was too low
-		icon_state = "damage_2"
-		desc = "A cage for containing SCP-173. It is currently holding [current173]. It looks fairly damaged."
-	else if(damage_state > damage_max) //Damage state is equal to damage max last stage before breaking, so this is neccesary
-		icon_state = "damage_3"
-		desc = "An empty cage for containing SCP-173. It appears broken."
-
-	if((LAZYLEN(contents)) && current173)
-		underlays += image(current173)
+		return
```
TheDarkElites, line 405: This causes the cage sprite icon_state to remain "open" even when the cage is broken, as 173 is no longer inside and  if(!LAZYLEN(contents)) returns true.

TheDarkElites: you can ignore this though I'm sure youre aware of it

EgorDinamit:
Alright, changed.

PR #1049, opened 2023-04-10
adds codex info for 173 blink mechanics

TheDarkElites:
## About the Pull Request
Makes it so people can now figure out how to delay their blinks because that might be important.

## Why It's Good For The Game

Important info for players.

## Changelog

:cl:
add: 173 Codex entry
/:cl:


aidan-bowman:
Probably would be more fitting to have an article specifically on blinking and link it directly to the blink GUI thingy. I don't think we should have codex entries for SCPs, primarily because people are intended to actually read the documents in the archives.

TheDarkElites:
> Probably would be more fitting to have an article specifically on blinking and link it directly to the blink GUI thingy. I don't think we should have codex entries for SCPs, primarily because people are intended to actually read the documents in the archives.

Good point. Counter argument: I'm lazy.

TheDarkElites:
drafting this so I can do what cheese said later

TheDarkElites:
Hud elements return custom messages on examine and don't automatically include codex entries. I could make it included for blink specifically, but it feels too out of place so I'm closing this PR since there is no viable place to put the codex entry.

PR #1070, opened 2023-04-14
Mapping : 2427-3 HCZ transfer + Guard Slots Tweak + Armoury Changes (Stunbatons etc)

GGarciaECE:
## About the Pull Request
First, the PR lowers slots of LCZ guards to a reasonable level for our population level. Instead of 20 slots being available, there will be 8 LCZ guards slots + 2 Sergeants (total of 11 including the Zone Commander). Since there are two fewer sergeants, the sergeant role become somewhat more important and central to running the sector (similar to EZ) and the ZC will be able to rely on them more (having too many sergeants dilute the authority of the role)

For similar reason, HCZ Sergeant slots have been reduced to 2 from 4 (HCZ has a total number of slots of 9 now, just like EZ which hasn't been changed).

The PR also does mapping changes around the different security offices and armoury to reflect changes in slot number.
LCZ has had more changes than the others, with several weapons being removed (the M16s lockers notably) whilst some additional non lethal options added.

173 chamber has been slightly reworked to include a more compact custodial closet and a more secure 'airlock' portion where Class D can be sandwiched between two gates for transfer into the Containment chamber itself. Won't prevent a breach, but will prevent stupidity from dooming the site early on.

Remapped 2427-3 (spooder) chamber. Relocated to HCZ close to the entry checkpoint , its slightly more secure now with an airlock style blast gate system. Still has the delivery chute and pen for the animals.

## Why It's Good For The Game
Less LCZ slots is in line with the current population of the game (which overs at 40-50). Having 20 slots was excessive and more of a leftover from the old 150/+ pop time which we no longer experience. Also makes LCZ Security area less insane, with much fewer lockers.
More secure 173 Chamber will simply prevent extremely dumb incidents from occuring too much. Guard stupidity can still get them killed but they now have some more layer of protection. 173 can absolutely still breach however.
Stunbatons are in line with the technological level imo, at least more so than tasers. I wish we had better sprites for them tho.

Spider chamber was improved and relocated to HCZ. Why HCZ?
Because in my opinion, LCZ guards are already focused on the CD Containment zone + 173 cycle and adding yet another distraction wouldn't really help. Conversely, HCZ guards often have nothing to do, so giving them a repetitive task (feeding the spider with goats ordered from Cargo) will make them somewhat more busy.

## Changelog

:cl:
add: Switched number of guards slots (20->8 for LCZ Guards, 4-> for LCZ Sergeants, 3->2 for HCZ Sergeants)
add: Modified contents of security offices of various zones to reflect changes in slots
add: Modified content of Security armouries to include Stun batons. LCZ has had several weapons removed with focus on making less than lethal more accessible.
add: Remapped 2427-3 (spooder) chamber. Relocated to HCZ close to the entry checkpoint , its slightly more secure now with an airlock style blast gate system. Still has the delivery chute and pen for the animals.
/:cl:

### DIFF
--- a/maps/site53/job/jobs/security.dm
+++ b/maps/site53/job/jobs/security.dm
@@ -221,8 +221,8 @@
 	title = "LCZ Sergeant"
 	department = "Light Containment Personnel"
 	department_flag = SEC
-	total_positions = 4
-	spawn_positions = 4
+	total_positions = 2
+	spawn_positions = 2
 	balance_limited = TRUE
 	//duties = "<big><b>As the Guard you have more access than a Junior Guard, but do not control them. You are to guard tests and SCP's in the zone you spawned in. If in doubt, ask your Zone or Guard Commander. You also have the duty of guarding the D-Class Cell Blocks. You should not leave your zone under usual SoP.</b></big>"
 	economic_power = 4
@@ -273,8 +273,8 @@
 	title = "HCZ Sergeant"
 	department = "Heavy Containment Personnel"
 	department_flag = SEC
-	total_positions = 3
-	spawn_positions = 3
+	total_positions = 2
+	spawn_positions = 2
 	//duties = "<big><b>As the Guard you have more access than a Junior Guard, but do not control them. You are to guard tests and SCP's in the zone you spawned in. If in doubt, ask your Zone or Guard Commander. You also have the duty of guarding the D-Class Cell Blocks. You should not leave your zone under usual SoP.</b></big>"
 	economic_power = 4
 	alt_titles = list("HCZ Senior Containment Response Agent", "HCZ Containment Response Sergeant", "HCZ Senior Combat Medic", "HCZ Senior Agent")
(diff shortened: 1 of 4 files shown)


magicianofwands:
are you okay

GGarciaECE:
image

GGarciaECE:
> are you okay

no :(

EgorDinamit (changes requested):
The 2427-3 room is fine, but let's not add stun batons to the starting guard loadout, okay?

EgorDinamit (changes requested):
The 2427-3 room is fine, but let's not add stun batons to the starting guard loadout, okay?

GGarciaECE:
> The 2427-3 room is fine, but let's not add stun batons to the starting guard loadout, okay?

Why exactly?
We already had tasers/shock carbines which are way more OP than stunbatons ever could hope to be. 
We need a replacement for the telescopic batons, they suck (not because they are bad at combat but because beatings end up in trips to medbay every single times).

EgorDinamit (approved):
Alright, I forgot that those are not /tg/ stun batons.
Let's get it merged.

EgorDinamit (approved):
Alright, I forgot that those are not /tg/ stun batons.
Let's get it merged.

PR #1071, opened 2023-04-15
Remove broken gamemode from game panel

checkraisefold:
## About the Pull Request

Tin, removes the 'Random' gamemode from the game panel. It causes the server to instantly restart whenever the game starts with it selected, because there is literally no handling for it in SSticker. I didn't add the gamemode handling because it's useless; the only functional difference between it and Secret is that it would tell players the selected gamemode, which is frankly never useful.

## Why It's Good For The Game

Dawson restarted the server 3 times by accident by selecting it.

## Changelog

:cl:
admin: Removed 'Random' game mode from game panel.
/:cl:

### DIFF
--- a/code/modules/admin/topic.dm
+++ b/code/modules/admin/topic.dm
@@ -657,7 +657,6 @@
 		for(var/mode in config.modes)
 			dat += {"<A href='?src=\ref[src];c_mode2=[mode]'>[config.mode_names[mode]]</A><br>"}
 		dat += {"<A href='?src=\ref[src];c_mode2=secret'>Secret</A><br>"}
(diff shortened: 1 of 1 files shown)


EgorDinamit (commented):
This is really lazy...
I'd imagine we can trust admins to not use it until someone fixes it?

EgorDinamit (approved):

checkraisefold:
> This is really lazy... I'd imagine we can trust admins to not use it until someone fixes it?

Like I said, there's no point in adding it. I could add it in five minutes, except it literally has only a slight difference from Secret. There's no point.
No, Dawson already selected it, which restarted the server once, then *again* because of selection saving. Can't really trust the admins.

PR #1011, opened 2023-03-10
Ocular and  Audio Memetic Protection System + Blink Rework

TheDarkElites:
### Summary
This takes all of the fun and different ways that procs and other things have been checking to see if humans are blind and combine them into one neat function, can_see()! This also adds the same for audible objects, with can_hear(). Both also can be passed with a var to make it a memetic check, with specialist memetic insulation taking priority in said checks. Additionally, we now have can_identify(), which is a more thorough can_see() check for objects that you need to be able to identify, read, or otherwise examine rather than just have it visible to you. Finally, this reworks the blinking system to be relative to humans, not 173, and also reworks blinking. The variability of blinking times has decreased, and so have the assigned blinking time lengths. Now you must spam space in order to delay your blink.

### For Players

Now, this is mainly code focused, so the changes players will notice are minimal. The most drastic change is the new blinking system, which, as explained above, requires you to spam your space key (or a differently chosen hot key) to slow down how fast you blink. No longer are you a victim of chance; you now control your concentration! You also can now tell how much time you roughly have before you blink.

The next change you'll notice is that when putting on blindfolds, ear mufflers, or otherwise having your visibility or hearing ability impacted, the response from the in-game environment is now more uniform. Since I've gone through and put in the new functions almost everywhere they were deemed necessary, there should no longer be any weird discrepancies.

Finally, examining objects has been implemented with the new can_identify() proc. As such, being able to examine objects is now determined by distance, nearsightedness, and the size of the object. This also applies to mobs. Turfs and anything else listed remain unaffected.

### For our Esteemed Contributors

From now on, you should endeavor to use can_see() and can_hear() in your code. It can be used for all mobs, and can_see() and can_hear() replace isblind() and isdeaf(), respectively. As previously mentioned, can_see() and can_hear() are much more extensive and can be used as follows:

mob.can_see() checks if the mob can see in general, either from disabilities, clothing, or anything else.
mob.can_see(origin) checks if the mob can see origin, and this should be used whenever possible to ensure consistency.

mob.can_hear() checks if the mob can hear in general, either from disabilities, clothing, or anything else.
mob.can_hear(origin) checks if the mob can hear origin, and this should be used whenever possible to ensure consistency.

Both also have unique variables that can be passed with them to change their function, and you should refer to memetic_insulation.dm for documentation.

[...]

PR #1099, opened 2023-04-18
Fix repetitive hear() checks in hear_say

### LINKED ISSUE #1093: scp-2427-3 intercom no work
#### Description of issue
SCP-2427-3's intercom prolly doesn't work. When I played as the spider, I could hear other people and see my messages transmit, but nobody actually responded. 

The spider prolly shouldn't be ignored if they're gonna be in a cell all day.

#### Steps to reproduce

Be the spider and try having a conversation in your containment.

#### Specific information for locating

I dunno

Issue been happening since today atleast

#### Issue bingo

checkraisefold:
## About the Pull Request

For some reason, the is_deaf call in hear_say was replaced with a can_hear call in #1011. This is a horrible idea, because anything calling hear_say **already checks for people in range.** Essentially, the code was calling hear(7, get_turf(src)) twice per hear_say call, because the code calling hear_say already does this.

Review silicon say.dm and mob/living say.dm to verify.
Tested and working on a local server.

## Why It's Good For The Game

This change also broke several things, including being able to hear the AI when it talks through a holopad.
This may or may not fix #1093, it seems related to this change.
Also, `\him` wasn't working for some reason. I just ported the upstream Baystation fix for this issue, which changes it to 'says something you cannot hear'

## Changelog

:cl:
fix: The AI talking through holopad with :h now works correctly again.
spellcheck: 'talks but you cannot hear .' has been replaced with 'says something you cannot hear.'
/:cl:

### DIFF
--- a/code/modules/mob/hear_say.dm
+++ b/code/modules/mob/hear_say.dm
@@ -61,12 +61,12 @@
 		if(get_preference_value(/datum/client_preference/ghost_ears) == GLOB.PREF_ALL_SPEECH && (can_see(speaker)))
 			message = "<b>[message]</b>"
 
-	if(!can_hear(speaker) || get_sound_volume_multiplier() < 0.2)
+	if(is_deaf() || get_sound_volume_multiplier() < 0.2)
 		if(!language || !(language.flags & INNATE)) // INNATE is the flag for audible-emote-language, so we don't want to show an "x talks but you cannot hear them" message if it's set
 			if(speaker == src)
 				to_chat(src, SPAN_WARNING("You cannot hear yourself speak!"))
 			else if(can_see(speaker))
-				to_chat(src, "<span class='name'>[speaker_name]</span>[alt_name] talks but you cannot hear \him.")
+				to_chat(src, "<span class='name'>[speaker_name]</span>[alt_name] says something you cannot hear.")
 	else
 		if (language)
 			var/nverb = verb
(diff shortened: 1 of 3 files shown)


code/modules/mob/hear_say.dm
```diff
@@ -61,12 +61,12 @@
 		if(get_preference_value(/datum/client_preference/ghost_ears) == GLOB.PREF_ALL_SPEECH && (can_see(speaker)))
 			message = "<b>[message]</b>"
 
-	if(!can_hear(speaker) || get_sound_volume_multiplier() < 0.2)
```
TheDarkElites, line 64: This will break imperfect audio checks as it relies on the distance to origin to decide if you can hear with imperfect audio insul. Additionally, although, as of right now, can_hear() only uses origin to check distance based on audio insulation, and this only matters in imperfect audio insulation, in the future, there may be adjustments to the memetics system that will require the origin. In essence, the origin should always be passed when checking if you can hear something, not if you can just hear; otherwise, bugs will ensue. In this case, the situation calls for a check of if you can hear something, not just hear in general. This will fix one bug and re-open another one while laying the foundation for multiple possible bugs in the future. I strongly encourage you to reconsider taking this approach, and if you have any questions, please I would be happy to VC with you to discuss. Otherwise, I've left you a mention on Discord that basically says everything I say here.

checkraisefold: Please give me an alternative approach to fixing holopads that does not require adding an extra argument to hear_say and can_hear that bypasses your range checks.

Also, why are you doing this in hear_say and not in say.dm files where hear_say is called? This would avoid the problem entirely, as custom audio insulation behavior can be done for different mobs.

checkraisefold:
agreed on better fix with @TheDarkElites

PR #1114, opened 2023-04-20
Fax Machine Limitations-Incentivizing fax communication.

TerryTehCoder:
Adds in another fax machine that's only accessible from the communications tower. This is the Only fax machine that can make outbound communications to command and other sites like the GOC, Horizon Initiative, etc. 

-Fax Machines no longer default to the Foundation Central Office and instead default to a random department.

-Two types of fax machines, the relay, and regular. The relay can send messages to command and others, default fax machines cannot.

## Why It's Good For The Game

A few folks in the discord suggested this and I quite like the idea. The hope is that when you force people to relay through one single machine they'll have to first send their communications over to the comms tower, and then they'll have to wait for communications to send their message out. Then they'll have to wait for communications to receive a new fax and send it back to their office. Ideally this would leave room for possible sabotage in between, a game of telephone, and give extra importance to the communication personnel's work. 

Example: 
1. GOC REP writes a paper
2. Paper is sent to Communications
3. Communications sends the paper out to its destination
4. Communications receives a response to the paper
5. Communications sends the response back to the GOC REP. 

While the premise sounds much like you're just adding waiting to the cycle of communications I do think it has some merit. The examples cited in conversation were running to the communications tower to send an emergency message when the communications staff were dead, or having a staff member able to fix up errors in the documentation.

As a reminder: **Administrative staff (server) can still _SEND_ a fax to _any_ fax machine they want with the Send Fax verb, however a warning is included to encourage sending secure communications to the communications tower.**

I am open to feedback on the topic, I suggest this is kept open for a little bit to leave room for changes.

## Changelog

:cl:
add: New Fax Machine Subtype, with regular and outbound destinations.
add: Outbound Fax machine to the communications tower
add: Communications Briefing about the new fax relay.
changed: Normal Fax machines no longer default to the foundation central office.
changed: Normal fax machines can no longer contact off-site locations.
/:cl:


checkraisefold:
this sounds like a horrible idea that will turn comms officer/comms dispatcher (which doesnt even have access to the fax by default) into a job of sitting waiting in the office for faxes, and if they're not sitting in there waiting, then faxes take like 8 years to get to admins or never get sent at all

people already hardly send faxes, this is a good way to make them send none

aidan-bowman:
yeah tbf this would only work if comms officer was guaranteed every round, do you think you could make this functionality dependent on playercount/comms officer existing?

Vizoplez:
This is a horrible idea balance-wise too. Comms Officer/techs can be antagonists and just secretly refuse to send any faxes and blame it on admins being slow. From the 'balance guy', this gets a hard NO, and was not discussed with any developer/maintainer before being submitted.

This also removes the layer of 'GOC/MC&D/ETC' are all their own factions, why in the name of hell would they let a comms officer read what they've written???

This is just a hard NO. From a balance & staff perspective

TerryTehCoder:
I honestly agree with the points brought up. I'm gonna close this for now. I don't think having a check for comms staff would matter much on the subject either. I do think communications still needs some major insights but this is not the way to go about it.

Next time I'll rope some more discussion into it beforehand, my apologies.

PR #1116, opened 2023-04-21
173 AI Rework

TheDarkElites:
## About the Pull Request

Basically, uh 173's AI kinda mid and cant take advantage of 173's new abilities. Also it just teleports behind people so kinda oof.

173 Can now attempt to free itself from its cage
173 No longer just teleports behind people but instead moves towards them (very quickly)
173 Can now properly wander
173 Now remembers targets and can track them through doors
173 Can now pathfind through obstacles he can destroy (windows and doors)
173 Will now attempt to flee when facing more than two humans
173 Will now attempt to break lights when appropriate

This merge also contains  #1123
## Why It's Good For The Game

Better AI good (I'm going to make his AI enabled in a later PR even when in low pop and just make it so he cant breach himself and cant be controlled by a human instead of just being catatonic under 30 pop)

Uhhh oh yeah hold on I had something else here

_flips through notes_

yeah, _ahem_, fuck you cheese stay away from my manual email password.

That is all.

## Changelog

:cl:
add: 173 AI now good :)
balance: 173 is now active in low-pop but AI only and cannot self-breach.
/:cl:


TheDarkElites:
time to get yelled at for shitcode :')

Tsurupeta:
Arguably, teleporting behind is better than attempted pathfinding via walking...

TheDarkElites:
> Arguably, teleporting behind is better than attempted pathfinding via walking...

No? Only maybe in terms of cost.

A.) Teleporting in the previous system means that if you were in 173's view, his snap cooldown was expired, and you blinked, there was a 100% chance of death. This basically makes 173 incredibly infuriating and incredibly difficult to counter from as no matter what you do if you can't evade him if you're backed into a corner. This way, there is a chance that 173 might not catch up to you or otherwise gives players the ability to evade/outrun him better if they're being smart rather than just luck of whether or not your blink timer runs out or if you can find a door.

B.) This isn't pathfinding via walking. This is pathfinding and then following that path. This is what allows 173 to follow players through objects as it lets him actually break open doors and follow through. Before 173 would basically sit in whatever room he was in while moving very little until someone found him. Then, as long as you closed a door on him, you were fine. Additionally, you cant replicate this pathfinding through obstacles with the previous system unless you're fine with 173 teleporting through doors.

In essence, this encouraged the gameplay loop of "find a door or get out of sight before blink or die". This means in situations where you weren't backed into a corner, its piss easy to escape, but in situations where you are cornered, it's instant death no matter what you do.

This evens out that gameplay loop by first stopping doors from being a be all end all escape route and actually allowing the AI to overcome obstacles (like doors/windows) but also giving people backed into a corner a chance to escape by using blinking as a method to lure 173 away from chokepoints (much like in CB). 

TLDR:

Its a little more expensive, sure, but it majorly improves gameplay against the 173 AI.

PS: We need competent AI that can simulate an experience similar to if a player was controlling them for non-RP hostile SCPs since we are likely to make such SCPs only playable on breach, so in all other cases, the SCPs need to be still able to provide fun and balanced gameplay when only controlled by their AI.

TheDarkElites:
Also please explain what you mean by "attempted pathfinding"?

TheDarkElites:
do not TM, fixing issues

TheDarkElites:
Okay, the issues Tsuru brought up about 173 being easy to doge have been addressed

TheDarkElites:
ready for TM whenever

code/modules/SCP/SCPs/SCP-173.dm
```diff
@@ -321,7 +301,150 @@ GLOBAL_LIST_EMPTY(scp173s)
...
+						assign_target(pick_turf_in_range(loc, flee_distance, list(/proc/isfloor)))
+						if(world.time > while_timeout)
+							break
+				else
+					assign_target(get_viable_light_target())
+			else
+				assign_target(DEFAULTPICK(possible_human_targets, null))
+
+	move_to_target()
+	return
```
ivanmixo: Pointless return

code/modules/SCP/SCPs/SCP-173.dm
```diff
@@ -321,7 +301,150 @@ GLOBAL_LIST_EMPTY(scp173s)
...
+	if(!new_target)
+		return FALSE
+
+	var/list/temp_steps_to_target = AStar(loc, get_turf(new_target), /turf/proc/AdjacentTurfsWithWhitelist, /turf/proc/Distance, max_nodes=flee_distance * 2, max_node_depth=15, min_target_dist = 1, adjacent_arg = list(/obj/structure/window, /obj/machinery/door, /obj/structure/grille)) //Flee distance is used as max_nodes since that should be the farthest that 173's AI will ever attempt to path
+	if(temp_steps_to_target) //Double check to ensure that whatever target we assign we can actually get to
+		steps_to_target = temp_steps_to_target
+		target = new_target
+		target_pos_last = get_turf(new_target)
+		return TRUE
+	else
```
ivanmixo: Pointless else

code/modules/SCP/SCPs/SCP-173.dm
```diff
@@ -321,7 +301,150 @@ GLOBAL_LIST_EMPTY(scp173s)
...
+	if(istype(loc, /obj/structure/scp173_cage))
+		loc.relaymove(src, NORTH)
+		return
+
+	var/list/possible_human_targets = list()
+
+	for(var/mob/living/carbon/human/H in dview(14, src)) //Identifies possible human targets. Range is double regular view to allow 173 to pursue tarets outside of world.view to make evading him harder.
+		if(H.SCP)
+			continue
+		if(H.stat == DEAD)
```
ivanmixo: Could prob combine these two into an OR. AStar one can stay separate since it's huge.

ivanmixo (changes requested):

ivanmixo:
Can't believe we don't have an alive human list though. TODO

ivanmixo (approved):

PR #1125, opened 2023-04-23
Fix show today server log

checkraisefold:
## About the Pull Request

there is an issue where if its past 12am GMT (the live server's timezone) while the current round is running, show today's server log will fail and nobody will be able to get current round log past this time

this changes it to use diary, which is the current day's log defined for the current round that's ongoing, the next day's won't exist until the next round

## Why It's Good For The Game

this is so annoying when staffing the server at 4pm my time on 40 players

## Changelog

:cl:
admin: Show today server log works correctly after a current round passes 12am server timezone
/:cl:

### DIFF
--- a/code/modules/admin/verbs/getlogs.dm
+++ b/code/modules/admin/verbs/getlogs.dm
@@ -91,11 +91,10 @@
 	if(!check_rights(R_ADMIN|R_MOD, TRUE))
 		return
 
-	var/path = "data/logs/[time2text(world.realtime,"YYYY/MM/DD")].log"
-	if( fexists(path) )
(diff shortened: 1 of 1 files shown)


ivanmixo (commented):
I hate the spaces in the if statement. Tested this?

checkraisefold:
> I hate the spaces in the if statement. Tested this?

Spaces in the if statement are there for consistency with rest of the file unfortunately, I hate it as well. Could use a big regex replace pr in the future to fix this issue permanently.
Otherwise, this is tested and working locally. Good to merge.

ivanmixo (approved):

PR #1136, opened 2023-04-28
More codex shit

aidan-bowman:
## About the Pull Request

Just a huge as hell PR to improve and build on the codex.

## Why It's Good For The Game

A large amount of these changes made the Codex more wiki-like and expandable.

## Changelog

:cl:
add: New codex categories and entries.
qol: Codex restrictions (i.e. cooldown system and mob-based access) have been removed,
fix: Massive codex improvements.
/:cl:


TheDarkElites (approved):
looks fine to me but I would like it to be looked over by tsurpta aswell before its merged.

Will add into today's tm.

code/modules/clothing/head/misc.dm
```diff
@@ -196,31 +214,37 @@
 	icon_state = "bowler_hat"
 	desc = "For the gentleman of distinction."
 	body_parts_covered = 0
+	hidden_from_codex = FALSE
```
Tsurupeta, line 217: why do you need to type "false" in every item if you can just do some funny things with abstract types so they're always hidden from codex??

PR #1151, opened 2023-05-05
Abominable infestation port

EgorDinamit:
## About the Pull Request

Ports the following PRs:
- #vlggms/tegustation-bay12/pull/172
- #vlggms/tegustation-bay12/pull/351
- #vlggms/tegustation-bay12/pull/374

Description from ported PR:
`A very dangerous and somewhat hard to deal with infestation of unknown alien species that evolve at a rapid pace.
Some planets may be completely taken over by it at the start of the round(basically, new exoplanet type), requiring exploration crew to salvage what is still left and order orbital bombardment once it's done.
Some planets might only be on the initial stage of infestation(random ruin with 2-3 mobs) which can be stopped.
If you are extremely unlucky - one of larvas might find its way on the ship. What happens from that point is mystery.`

Here, the event is not enabled just yet, nor do we have exoplanets, so that part was omitted.
What was ported is the mobs themselves, several simple mob improvements(death sounds, burst fire capabilities, etc), the riot officer simple mob(the only one that uses burst fire at the moment) and associated chemicals.

This is no more than an admin thing at the moment, as nothing can normally trigger the event, but it might change in the future.

## Changelog

:cl:

add: Abominable Infestation: The infestation mobs can potentially find their way onto the facility via vents network. They evolve rapidly and "disassemble" other creatures to produce more of its kind. Their meat contains potentially powerful chemicals, but be careful not to consume it raw.
/:cl:


Tsurupeta:
the sprites are sooooooooo... Eye-catching..............

77seven77seven:
Could we possibly try to rename this to SCP-610?

EgorDinamit:
> the sprites are sooooooooo... Eye-catching..............

I am a very good spriter :)
And yeah, we need proper sprites for assembler, eviscerator and the structures.

Tsurupeta (approved):

PR #1161, opened 2023-05-15
Ammo optimization

Tsurupeta:
Bullets are now created when shooting gun instead of being stored inside ammo_casing.

P90 mag now holds only 50 atoms inside instead of 100

Speeds up init time on my machine by ~5 seconds


Tsurupeta:
Testing showed positive results.

PR #1282, opened 2023-08-21
Deletes botany and piloting skills

aidan-bowman:
## About the Pull Request

Removes botany and piloting skills. A few piloting skill references in mechs were moved to weapons skill, all other skill-checks were wiped

## Why It's Good For The Game

The skills system is intended to solve one major problem: players powergaming by being able to use different department-specific systems (e.g. doing surgery right after turning on the reactor) with no drawbacks or consequences. However, our current system is downright trash at this. Most jobs get decent pre-defined skills in their intended jobs AS WELL AS extremely high skill points that they can spread over practically every skill they could ever want. This, coupled with the fact that skills are hilariously unbalances, doesn't solve and in fact WORSENS the issue of powergaming.

The ideal solution would be to simplify the system as much as possible, hopefully to the point of TG where it's basically either "you have the role or you dont", but as this is a major system that practically everyone interacts with, I'd prefer to make slow, gradual changes until we achieve the balance we want. Removing botany and piloting is just low hanging fruit; there's no specific botany role to "protect" (the cook and the bartender are currently expected to manage botany as well, although hopefully this would be replaced. other than that, the CDZ botany is practically never used except to metagame baseball bats), and piloting isn't even something we HAVE (I guess you could maaybe use it for driving/piloting vehicles, but that's so specific I don't think that needs its own skill).

## Changelog

:cl:
del: Removed botany skill. All related skill checks have been removed.
del: Removed piloting skill. Mech piloting checks have been transferred to weapons skill, and other references have been removed.
/:cl:


TheDarkElites:
botany skill does do stuff

TheDarkElites (approved):
meh

TheDarkElites:
Technically test merged once but only for a short time so will be slotted for next tm

PR #1300, opened 2023-08-26
Adds a burn chamber inside GOC base

ChuChu42:
## About the Pull Request

adds a burn chamber in GOC base, behind the armory
image

## Why It's Good For The Game

GOC are cannonically known for capturing then incenerating SCPs, this chamber gives them a place to incenerate said NPCs, tough being basically an atmospheric plasma burn chamber, it still works

## Changelog

:cl:
add: Plasma burn chamber in GOC base
/:cl:

### DIFF
--- a/maps/site53/z1_admin.dmm
+++ b/maps/site53/z1_admin.dmm
@@ -17,6 +17,12 @@
 /obj/effect/wallframe_spawn/reinforced,
 /turf/simulated/floor/shuttle/black,
 /area/site53/tram/mtf)
+"ap" = (
+/obj/machinery/igniter{
+	id_tag = "GOC"
+	},
+/turf/simulated/floor/reinforced/oxygen,
+/area/centcom/goc)
 "ar" = (
 /obj/effect/floor_decal/corner/paleblue/bordercorner{
 	dir = 1
@@ -227,6 +233,13 @@
 	},
 /turf/simulated/floor/shuttle/white,
 /area/site53/tram/mtf)
+"ct" = (
+/obj/machinery/atmospherics/portables_connector{
+	dir = 8
+	},
+/obj/machinery/portable_atmospherics/canister/phoron,
+/turf/simulated/floor/tiled/techfloor,
+/area/centcom/goc)
 "cu" = (
 /obj/effect/floor_decal/industrial/hatch/yellow,
 /obj/machinery/door/blast/shutters{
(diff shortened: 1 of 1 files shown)


aidan-bowman:
First off, re-write the changelog

Second off, I can't speak for the administration, but I do think it'd be better to go in a direction OTHER than the series 1 "kill everything that moves" GOC, it's poorly written, uninteresting, and frankly makes no fucking sense

ChuChu42:
> First off, re-write the changelog
> 
> Second off, I can't speak for the administration, but I do think it'd be better to go in a direction OTHER than the series 1 "kill everything that moves" GOC, it's poorly written, uninteresting, and frankly makes no fucking sense

i dont fully expect it to be merged i simply added it because i think its a good idea to give GOC an actual incenerator to try burning SCPs (and fail to do so) after capturing them as much futile it is

ghost:
> First off, re-write the changelog
> 
> Second off, I can't speak for the administration, but I do think it'd be better to go in a direction OTHER than the series 1 "kill everything that moves" GOC, it's poorly written, uninteresting, and frankly makes no fucking sense

The burn chamber is not to "kill everything that moves", i don't think it will be able to kill scp's and it's just a tool, there is no obligation to use the tool

checkraisefold:
this is decent, but what if instead of a burn chamber we add some other tool that the GOC can use to liquidate parathreats? this would make more sense in lore - the only TE's they'd really be liquidating with a burn chamber would be like type reds and some other things but it's not very versatile to deploy in the field.

maybe add a gun that does much higher damage to anomalies or something? idk, something to come up with that isnt just plasma oxy generic burn chamber mapping change

aidan-bowman:
> > First off, re-write the changelog
> > Second off, I can't speak for the administration, but I do think it'd be better to go in a direction OTHER than the series 1 "kill everything that moves" GOC, it's poorly written, uninteresting, and frankly makes no fucking sense
> 
> The burn chamber is not to "kill everything that moves", i don't think it will be able to kill scp's and it's just a tool, there is no obligation to use the tool

So? Should we give LCZ guards a "tool" to kill every D-class? Should we give traitors a "tool" to immediately explode the nuke? If something doesn't fit the vision we have for the game it's not getting merged, simple as.

BadAtThisGame302:
Can you map a UIU base? It's the only GOI apart from Marshals (Which is like a bar or something) that doesn't get a base or available gear for a raid on the Site. (In another PR, not this one.)

ChuChu42:
closed, shitpost PR all along anyways

PR #1301, opened 2023-08-26
Remove the Ahelps to IRC notice.

Tuulipommi:
## About the Pull Request

Removes the Red text saying Ahelps are sent to IRC when pressing Staffwho, since it does not do that.

## Why It's Good For The Game

Removes a lie from the server.

## Changelog

del: Removed Ahelps to IRC note.

/:cl:

### DIFF
--- a/code/game/verbs/who.dm
+++ b/code/game/verbs/who.dm
@@ -107,7 +107,5 @@
 		else
 			msg += line
 
-	if(config.admin_irc)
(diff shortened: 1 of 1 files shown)


aidan-bowman:
Why is `config.admin_irc` set if we don't actually have an IRC? also, wouldn't it be better to actually add that functionality?

Tuulipommi:
> Why is `config.admin_irc` set if we don't actually have an IRC? also, wouldn't it be better to actually add that functionality? 

Remnant of old code. And maybe later, not a priority atm.

TheDarkElites (approved):
Looks fine enough

PR #1356, opened 2023-10-27
Adds "succumb" verb to humans

EgorDinamit:
## About the Pull Request

Ports vlggms/tegustation-bay12/pull/364

- Adds "succumb" verb to IC tab for human mobs.

## Why It's Good For The Game

- Allows people stuck in near-death state to simply die. Sometimes, with Bay med, it takes too long to actually die, despite there being absolutely no chance of you surviving it even if you lie there for 10 minutes. Additionally, many people would already simply ghost in such scenarios, which is less than ideal in my opinion.

## Changelog

:cl:
tweak: Players can now succumb if their heart stops(no pulse), dying on the spot, allowing them to skip the grueling time of observing the black screen as they die for the next 10 minutes.
/:cl:

### DIFF
--- a/code/modules/mob/living/carbon/human/death.dm
+++ b/code/modules/mob/living/carbon/human/death.dm
@@ -106,3 +106,22 @@
 		E.status |= ORGAN_DISFIGURED
 	update_body(1)
 	return
+
+/mob/living/carbon/human/verb/succumb()
+	set category = "IC"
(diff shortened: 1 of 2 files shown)


EgorDinamit:
Succumb verb already existed, but it was rather useless as it only worked with dead heart and had no confirmation(imagine accidentally misclicking).

PR #1378, opened 2023-11-06
Rejuvenating cyborg give charges

Saicchi:
## About the Pull Request

Add recharge for Robots Rejuvenate Command

## Why It's Good For The Game

Makes rejuvenating a cyborg with no charges get some.

## Changelog

:cl:
fix: Rejuvenating robots now gives them charge
/:cl:

### DIFF
--- a/code/modules/mob/living/silicon/robot/robot_damage.dm
+++ b/code/modules/mob/living/silicon/robot/robot_damage.dm
@@ -34,8 +34,8 @@
 
 /mob/living/silicon/robot/rejuvenate()
 	..()
-	if(src.cell)
-		src.cell.give(src.cell.maxcharge)
+	if(cell)
+		cell.give(cell.maxcharge)
 
 /mob/living/silicon/robot/proc/get_damaged_components(brute, burn, destroyed = 0)
 	var/list/datum/robot_component/parts = list()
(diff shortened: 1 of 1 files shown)


77seven77seven:
a small but epic change

code/modules/mob/living/silicon/robot/robot_damage.dm
```diff
@@ -32,6 +32,11 @@
 	else
 		heal_overall_damage(0, -amount)
 
+/mob/living/silicon/robot/rejuvenate()
+	..()
+	if(src.cell)
+		src.cell.give(src.cell.maxcharge)
```
EgorDinamit: `src` is obsolete here.
```suggestion
	if(cell)
		cell.give(cell.maxcharge)
```

Saicchi: I'm a big fan of 'Explicit is better than implicit.' I'd rather not remove it unless it clashes with the code guidelines.

EgorDinamit: I am leaving the suggestion here for a reason. There is 0 reasons to utilize `src.var` outside of situations where you have a proc's argument named the same way as a variable of src object (i.e. `src.thing = thing`).

Tsurupeta: Just remove it

EgorDinamit (approved):

PR #1422, opened 2024-01-01
Re-Adds the ability for D-Classes to have their own names

ChuChu42:
## About the Pull Request

comments out the line that makes the dclass names randomize on spawn

## Why It's Good For The Game

i can understand developers and hosts wanting to make it so d-classes are randomly named on spawn
but it generally stops people from having unique characters as d-classes

this PR re-adds the ability for d-classes to use their own names

i really do not get why it was made to randomize names on spawn in the first place, it just stops people from having their own uniquely named, unique characters, making the player a entire diffrent person everytime they become a class-D, unable to disable this randomizing process

## Changelog

:cl:
code: comments out the d-class name randomizer
/:cl:

### DIFF
--- a/code/game/jobs/job/misc.dm
+++ b/code/game/jobs/job/misc.dm
@@ -24,7 +24,7 @@
 	duties = "You're a prisoner. You have no duties!"
 
 /datum/job/classd/equip(mob/living/carbon/human/H)
-	H.fully_replace_character_name(random_name(H.gender, H.species.name))
+//	H.fully_replace_character_name(random_name(H.gender, H.species.name))
 	. = ..()
 	var/r = rand(100,9000)
 	while (used_numbers.Find(r))


Teragenic:
You are a criminal and a disposable asset. I am against this

ChuChu42:
> You are a criminal and a disposable asset. I am against this

does not mean people cannot have their unique characters, sure, you can name them d-2331 but i really do not like being unable to make a unique character at all

checkraisefold:
no thanks I dont want to moderate d-class metagrudging

ChuChu42:
> no thanks I dont want to moderate d-class metagrudging

d-classes already metagrudg guards.

projectkepler-ru:
> d-classes already metagrudg guards.

do you have any evidence to back this up

TheDarkElites:
This exists primarily for immersion purposes, and for the most part, everyone in the foundation except other members of the D-class would refer to their classification number. However, I do get your point about characters not really feeling like they persist round to round due to the name change. D-Class won't be allowed to have regular names as, again, that would break immersion for a lot of individuals, but if this PR changes to make it instead so that the names selected by players playing as D-Class could only consist of numbers, that way we maintain immersion while allowing characters to persist, then this would be an acceptable PR. Otherwise, this won't be merged. 

A minor thing in addition to what I said: Anything can be a name, and therefore, I see no issue with numbers being a name.

ChuChu42:
> This exists primarily for immersion purposes, and for the most part, everyone in the foundation except other members of the D-class would refer to their classification number. However, I do get your point about characters not really feeling like they persist round to round due to the name change. D-Class won't be allowed to have regular names as, again, that would break immersion for a lot of individuals, but if this PR changes to make it instead so that the names selected by players playing as D-Class could only consist of numbers, that way we maintain immersion while allowing characters to persist, then this would be an acceptable PR. Otherwise, this won't be merged.
> 
> A minor thing in addition to what I said: Anything can be a name, and therefore, I see no issue with numbers being a name.

still i think it would be more enjoyable to make players able to name themselves in order to make interesting characters

we can take APE MAN the d-class for an example

it is possible to make it only numbers, however, i do not know how to code that 👍

TheDarkElites:
Im closing this because of my previously stated reason and the fact that the proper solution for this is something that @K4rlox cant currently implement. It probably should be revisited in the future with the proper solution, given that I do think we should allow D-Class to have saved numerical names.

PR #1439, opened 2024-01-04
Ossarepantes tweak

### LINKED ISSUE #1438: ossarepantes can explode hands and feet
#### Description of issue

image
it can create an infinite cycle of restoring the limb then having it re-break for some reason instantly and then continues to keep restoring the limb and it breaks in a cycle that creates 9 zillion brute damage. limb then explodes from brute damage horrific fortune

ossarepantes should not explode limbs

EgorDinamit:
## About the Pull Request

Closes #1438 

- Damage done to the limb reduced to 20, from 35.
- Damage from fixing a bone is done *before* bone is restored.

## Why It's Good For The Game

- Ossarepantes was initially balanced for Tegu, where limbs have nearly twice the health. This makes it less dangerous.
- Prevents people from being locked into eternal loop of suffering (Fix bone - Deal damage - Bone breaks due to damage - REPEAT AD INFINITUM).

## Changelog

:cl:
tweak: Ossarepantes no longer locks you into a loop of bone-breaking. It is now safe to use for its purpose.
/:cl:

### DIFF
--- a/code/modules/reagents/Chemistry-Reagents/medicines/medicines_core.dm
+++ b/code/modules/reagents/Chemistry-Reagents/medicines/medicines_core.dm
@@ -285,8 +285,8 @@
 	var/mob/living/carbon/human/H = M
 	for(var/obj/item/organ/external/E in H.organs)
 		if(E.status & ORGAN_BROKEN)
-			E.status &= ~ORGAN_BROKEN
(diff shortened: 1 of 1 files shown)


checkraisefold (commented):
i approve merge instant

PR #1446, opened 2024-01-04
Fixes 2427 stealthbug

aidan-bowman:
## About the Pull Request

Slightly shrinks the size of 2427's chamber, preventing powergamers from abusing the hunger system by standing juuuust out of range of corpses. I swear we had issues with this previously, apparently someone just suppressed the bug instead of fixing it.

## Why It's Good For The Game

Stealthbugs are bad, and maintainers shouldn't be expected to be fully aware of every single potential problem at all times. This'll stop some clueless mapper or coder from accidentally surfacing a major exploit.

## Changelog

:cl:
fix: Slightly reduced 2427's chamber size, fixing a potential stealthbug.
/:cl:


EgorDinamit (changes requested):
There is not a single bug this fixes or pretends to fix.

EgorDinamit:
Oh, also this PR changes some stuff in LCZ, which isn't mentioned in the description at all.

PR #1477, opened 2024-01-21
Adds scp-1507

Coxswain-Navigator:
## About the Pull Request

Adds a new SCP, SCP-1507

- Behaves similarly to a wild animal, attacks in a group when hostile. It also has an aggression range of about 2 tiles.

## Why It's Good For The Game
New SCPs are good I guess

## Changelog

:cl:
add: SCP-1507
/:cl:


code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
+/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507
+	name = "pink flamingo"
+	desc = "A ping flamingo that acts like a real one."
```
EgorDinamit: Aren't they plastic flamingos? Might want to clarify.
```suggestion
	desc = "A pink plastic flamingo that acts like a real one."
```

code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
...
+
+	ai_holder_type = /datum/ai_holder/simple_animal/retaliate/cooperative/scp1507
+	say_list_type = /datum/say_list/scp1507
+	natural_weapon = /obj/item/natural_weapon/scp1507
+
+/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507/Initialize()
+	. = ..()
+	SCP = new /datum/scp(
+		src, // Ref to actual SCP atom
+		"pink flamingo", //Name (Should not be the scp desg, more like what it can be described as to viewers)
```
EgorDinamit: Same here.
```suggestion
		"pink plastic flamingo", //Name (Should not be the scp desg, more like what it can be described as to viewers)
```

code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
...
+		SCP_EUCLID, //Obj Class
+		"1507" //Numerical Designation
```
EgorDinamit: Worth mentioning, but we might want to have to count the amount of flamingos, and name them appropriately, i.e. SCP 1507-1, SCP 1507-2 and so on.

code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
...
+	)
+
+/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507/Life()
+	. = ..()
+	if(resting)
+		update_icon()
```
EgorDinamit: Updating icon every Life() tick is generally not a good idea. See how 2427-3 handles "sleeping"

code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
...
+
+/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507/on_update_icon()
+	if(stat == DEAD)
+		icon_state = icon_dead
+	else if(resting)
+		icon_state = icon_resting
+
+/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507/death(gibbed, deathmessage, show_dead_message)
+	. = ..()
+	update_icon()
```
EgorDinamit: Not needed, death proc already sets the icon state to icon_dead.

code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
...
+
+/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507/proc/enrage(potency)
```
EgorDinamit:
```suggestion
/mob/living/simple_animal/hostile/retaliate/cooperative/scp1507/proc/Enrage(potency)
```
Do the same for all new procs while we're at it.

code/modules/SCP/SCPs/SCP-1507.dm
```diff
@@ -0,0 +1,68 @@
...
+	name = "pink plastic"
+	gender = PLURAL
+	attack_verb = list("smacked around")
+	force = 4
+	damtype = BRUTE
+	canremove = FALSE
+
+/datum/say_list/scp1507
+	emote_hear = list("flaps its wings")
+	emote_see = list("flaps its wings", "scratches the ground")
```
EgorDinamit: Those should be moved below initial mob's definition.

code/modules/SCP/SCPs/SCP-1507.dm
aidan-bowman: > Do the same for all new procs while we're at it.

no one uses this typecase other than you lol

code/modules/SCP/SCPs/SCP-1507.dm
aidan-bowman: you should probably use signals for this

aidan-bowman:
would be funny if there were regular lawn flamingos as well, and you couldn't tell the difference between the two if the skip stood still

aidan-bowman (changes requested):
This is missing an in-game SCP file.

EgorDinamit (approved):

PR #1501, opened 2024-02-26
Adds SCP-5295 + shitload of modcomp fixes

aidan-bowman:
## About the Pull Request

Implements [SCP-5295](https://scp-wiki.wikidot.com/scp-5295). This program lets you remotely steal files off of other computers once their owner deactivates them.

Also fixes portable hard drives (and places a few of them in the IT department's room)
Also fixes program menu icons.
Also adds a button to automatically perform a diagnostic check in the systems configuration program.

Some other code improvements were made.

## Why It's Good For The Game

New SCPs are good, new shit for the IT tech to worry about is good, things not being as broken is good.

## Changelog

:cl:
add: SCP-5295 is here! Located in LCZ, this program lets you remotely hack into other devices (if you can get their network ID). Slight mapping changes were made to accommodate the new skip.
add: The Computer Configuration program allows you to quickly scan internal components.
fix: Portable drives actually work now.
add: The IT department starts with a few portable drives.
fix: Programs have proper menu icons now.
/:cl:


code/modules/SCP/SCPs/SCP-5295.dm
TheDarkElites, line 1: Datum can't have an SCP ref associated with it obviously but I would like for you to have computers that are affected by this SCP to have a datum created for them, i.e "5295-1" type style.

TheDarkElites: Actually you can just have it be computers that download this get the datum, no need to apply it to those which are merely affected by the stealing properties of the computer because that might just spam the SCP list unnecesarily.

aidan-bowman: That sounds... awful. I thought the SCP ref was just for mobs we want people to spawn as?

TheDarkElites: SCP datum is meant to keep track of certain SCP related data (like flags for example) and keep track of scp items/affected items. So yes if a computer downloads an scp then the computer's SCP datum should be initialized.

As for why its awful, please elaborate?

aidan-bowman: > SCP datum is meant to keep track of certain SCP related data (like flags for example) and keep track of scp items/affected items. So yes if a computer downloads an scp then the computer's SCP datum should be initialized.

That sounds even worse! The SCP ref isn't a list, which means that something can only be affected by one anomaly at a time. Much more importantly, why are we even bothering to keep track of items affected by SCPs? I understand having a datum that tracks playable skips, but there's absolutely no reason for a var to work like that.

TheDarkElites: Again its not meant to represent purely affected, for example people who are affected by 012 dont have their scp datum initialized. Its just meant to track SCP atoms. In this case giving a computer an scp which which significantly impacts it would be justification.

Secondly the variable itself isint the method of tracking it just adds its scp atom to the scp list when initialized.

Thirdly, one way this is used is for when researchers or other individuals who should know about SCPs in game, examining an SCP gives them additional information identifying the SCP.

Again given these clarifications please explain how this is awful?

aidan-bowman: > Again its not meant to represent purely affected, for example people who are affected by 012 dont have their scp datum initialized. Its just meant to track SCP atoms. In this case giving a computer an scp which which significantly impacts it would be justification.

This skip HARDLY "significantly impacts" the computer it's on.
More importantly, I see no reason why we need to track atoms that are SCPs. As far as I'm concerned, the more similar "SCPs" are to regular atoms, the better.

> Secondly the variable itself isint the method of tracking it just adds its scp atom to the scp list when initialized.

I swear I've seen code that explicitly uses the SCP variable.

> Thirdly, one way this is used is for when researchers or other individuals who should know about SCPs in game, examining an SCP gives them additional information identifying the SCP.

brb examining some random guy's computer so i can identify every single program installed on it.

> Again given these clarifications please explain how this is awful?

I know you just coded up the whole SCP ref thing, and it stings to hear people deride it, but I just don't see a reason for it to exist. There are much better ways to solve all of the problems you might be trying to prepare for.


Research system involving SCPs? Unless we're going to have general-purpose experiments that work on _every single skip_ (awful idea), we can just use pre-existing TG experiment code.

Checking escaped/sold/stolen skips for effects? SCPs aren't the only things that are dangerous/valuable, and even if they were, this can be handled way better with a component.

Allowing ghosts to spawn in as SCPs? We can steal TG code for that too


Really, I just don't see any reason we would want an SCP ref, especially considering how much overhead one requires.

TheDarkElites: >I know you just coded up the whole SCP ref thing, and it stings to hear people deride it, but I just don't see a reason for it to exist. There are much better ways to solve all of the problems you might be trying to prepare for.

First, I would appreciate it if you would not assume that my decisions are based on any kind of bias towards my PRs. I would like to think that my position would make it clear I am capable of maintaining partiality. Secondly, I didn't "code in the SCP ref thing". Using an SCP var datum already existed before the SCP refactor, I just made it useful. 

>I swear I've seen code that explicitly uses the SCP variable.

As someone who coded it, I'm telling you the SCP variable doesn't do anything beyond data holding and some code when it's first initialized. It isn't touched after that. If you have a counter-example please show me instead of saying "I swear"

>More importantly, I see no reason why we need to track atoms that are SCPs. As far as I'm concerned, the more similar "SCPs" are to regular atoms, the better.

Sure, but we still need to track SCP-specific data for most SCPS, and I would rather be consistent than only have it initialized for some SCPs.

>brb examining some random guy's computer so i can identify every single program installed on it.

Yes because this is clearly what I meant. Actually look at the datum code first.

>especially considering how much overhead one requires.

What overhead? It's a single datum that only does any actual computation on initialization and then exists purely for data purposes. The best argument you could make for "overhead" here is how much space it takes up, which is barely anything. If initializing this datum caused any impact beyond the literal bare minimum, then your point would make sense, but it does not.

TheDarkElites: This aside, though, I do agree that the whole var thing is dumb; like I said, I didn't come up with it. It was already there. Maybe move it to a component in the future idk

TheDarkElites:
The SCP and the modcomp improvements should also be split into two PRs.

aidan-bowman:
> The SCP and the modcomp improvements should also be split into two PRs.

ooooh my god dude you are going to KILL me

TheDarkElites:
> > The SCP and the modcomp improvements should also be split into two PRs.
> 
> 
> 
> ooooh my god dude you are going to KILL me

Me when when cheese does not atomize PRs and then when told to atomize PRs dies

PR #1504, opened 2024-03-01
Macros improvements

aidan-bowman:
## About the Pull Request

Makes SOME improvements to macros:

Some end-of-line comments were moved to properly autodoc-ing comments.
Some flag defines were changed to use bitshifts.
(hopefully) ALL defines with unwrapped operators have been wrapped.

## Why It's Good For The Game

The first two improvements should be obvious.

An unwrapped define is something like `#define TEMPATURE_FOR_STUFF T20C + 10`
A wrapped define, in contrast, looks like `#define TEMPATURE_FOR_STUFF (T20C + 10)`

If you have something like `var/our_temp = TEMPATURE_FOR_STUFF * 2`, then this matters (because PEMDAS is a bitch):

With the former, the preprocessor gives you `var/our_temp = T20C + 10 * 2` which isn't what we expected.
With the latter, the preprocessor gives you `var/our_temp = (T20C + 10) * 2` which _is_ what we expected.

I'm mainly making this PR because apparently TG had [an issue](https://github.com/tgstation/tgstation/pull/37116) with this a while back and uhhhh no one here noticed.

## Changelog

:cl:
fix: A bug with some internal defines was fixed, potentially affecting some values (e.g. atmos)
/:cl:


code/modules/mechs/interface/screen_objects.dm
```diff
@@ -1,5 +1,5 @@
 // Screen objects hereon out.
-#define MECH_UI_STYLE(X) "<span style=\"font-family: 'Small Fonts'; -dm-text-outline: 1 black; font-size: 5px;\">" + X + "</span>"
+#define MECH_UI_STYLE(X) ("<span style=\"font-family: 'Small Fonts'; -dm-text-outline: 1 black; font-size: 5px;\">" + X + "</span>")
```
Tsurupeta, line 2: Why do you need this for string???

code/datums/uplink/uplink_sources.dm
```diff
@@ -1,4 +1,4 @@
-#define NO_GUARANTEE_NO_EXTRA_COST_DESC(X) "Installs an uplink into " + X + " if, and only if, found on your person. Has no TC cost."
+#define NO_GUARANTEE_NO_EXTRA_COST_DESC(X) ("Installs an uplink into " + X + " if, and only if, found on your person. Has no TC cost.")
```
Tsurupeta, line 1: Same here

code/modules/mechs/interface/screen_objects.dm
aidan-bowman: I guess it's not really important for strings... still, it's better to keep standards... standardized.

More importantly, this should probably just be a single string with a [x] inside of it.

aidan-bowman: > More importantly, this should probably just be a single string with a [x] inside of it.

Wait no I'm stupid

Tsurupeta (approved):

code/__defines/atmos.dm
```diff
@@ -25,7 +25,7 @@
 #define MAX_HIGH_PRESSURE_DAMAGE    4   // This used to be 20... I got this much random rage for some retarded decision by polymorph?! Polymorph now lies in a pool of blood with a katana jammed in his spleen. ~Errorage --PS: The katana did less than 20 damage to him :(
 #define LOW_PRESSURE_DAMAGE         0.6 // The amount of damage someone takes when in a low pressure area. (The pressure threshold is so low that it doesn't make sense to do any calculations, so it just applies this flat value).
 
-#define MINIMUM_PRESSURE_DIFFERENCE_TO_SUSPEND (MINIMUM_AIR_TO_SUSPEND*R_IDEAL_GAS_EQUATION*T20C)/CELL_VOLUME			// Minimum pressure difference between zones to suspend
+#define MINIMUM_PRESSURE_DIFFERENCE_TO_SUSPEND ((MINIMUM_AIR_TO_SUSPEND*R_IDEAL_GAS_EQUATION*T20C)/CELL_VOLUME)			// Minimum pressure difference between zones to suspend
```
TheDarkElites, line 28: unclosed parenthesis?

code/__defines/machinery.dm
```diff
@@ -2,9 +2,11 @@
 #define MEGAWATTS *1000000
 #define GIGAWATTS *1000000000
 
-#define MACHINERY_TICKRATE 2		// Tick rate for machinery in seconds. As it affects CELLRATE calculation it is kept as define here
+/// Tick rate for machinery in seconds. As it affects CELLRATE calculation it is kept as define here
+#define MACHINERY_TICKRATE 2
 
-#define CELLRATE (1 / ( 3600 / MACHINERY_TICKRATE )) // Multiplier for charge units. Converts cell charge units(watthours) to joules. Takes into consideration that our machinery ticks once per two seconds.
+/// Multiplier for charge units. Converts cell charge units(watthours) to joules. Takes into consideration that our machinery ticks once per two seconds.
+#define CELLRATE (MACHINERY_TICKRATE / 3600)
```
TheDarkElites, line 9: if this is intentional make sure you account for the fact you changed the cellrate from being a 1/ (3600 / MACHINERY_TICKRATE) to just  3600 / MACHINERY_TICKRATE.

code/modules/mob/observer/freelook/chunk.dm
```diff
@@ -1,4 +1,4 @@
-#define UPDATE_BUFFER 25 // 2.5 seconds
+#define UPDATE_BUFFER (2.5 SECONDS)
```
TheDarkElites, line 1: Make sure changing 25 // 2.5 SECONDS to 2.5 SECONDS is intentional

code/__defines/atmos.dm
TheDarkElites: im dyslexic

PR #1528, opened 2024-03-11
increases 012 damage since its veryyy boring playing the waiting game to succumb

pumpkintheproot:
## About the Pull Request

makes the 012 damage higher because frankly waiting there while you bleed out VERY slowly is VERY tedious and gets annoying, especially when you can't succumb (i know you can ghost but its probably preferred to not)
## Why It's Good For The Game
makes it so you dont have to wait 12057 years to actually die from 012's damage

## Changelog

:cl:
made 012's damage actually hurty owchie so now u can die faster
/:cl:


EgorDinamit (changes requested):
Do not just increase brute damage - this will simply put you to crit faster and potentially gib your hands.
Consider adding a timer OR make it cause heavy bloodloss by creating a wound in the hands (see how implanted larvas do it when they get in/out).

TheDarkElites:
Egor already mentioned this, but you can't increase brute damage because this means that a perfectly alive player will die faster instead of just succumbing faster.

Also, I'm pretty sure you can voluntarily succumb.

PR #1544, opened 2024-03-15
Ports Magic Rework from Tegu (476)

EgorDinamit:
## About the Pull Request

Full ports of vlggms/tegustation-bay12/pull/476

- All living mobs have mana level. Spells have varying mana costs, depending on their power. All "normal" people have it at 10, with regeneration rate of 0.25/s, whereas wizards have 100 and 2/s.
- Spell books changed entirely: Wizards are no longer forced to choose some "play style" and locking themselves into it, as the spell book they spawn with has *most* spells in it. When first used, an owner is selected, blocking anyone else from accessing it. A certain flag can make it available to anyone, instead. Dispelling the book will remove any locks.
- The mechanics of sacrificing items and "depositing" spell points removed.
- Spell points are no longer tied to spell books, instead, each mob has their own(in the mana datum), allowing anyone to learn magic, given they find a spell book that they can use.
- Added A LOT of new spells and updated the old ones.

There's probably a lot more..

## Why It's Good For The Game

I felt like wizard content was bland and boring, even when it was the "funny gamemode" - all wizard could do is pick some shitty spell book type and be forced to follow it. This is no longer a problem, as wizard can effectively pick any spell they wish to.
My intent with mana is to allow practically everyone to be able to use magic, while also limiting those that shouldn't be proficient in it. You still have to find a book somewhere, so it's not like it will be a common occurence.
The mechanics of getting more spell points via literally waiting for 30 minutes is garbage. Sacrificing items for them was a bit better, but I believe we should find a better way.

## Changelog

:cl:
add: Every human now has spell points and mana level. By default everyone has 2 points and 10 mana.
add: Spell books were reworked entirely.
add: Multiple new spells were added (around 20), with a lot of old ones(around 30) reworked and updated.
/:cl:


Vizoplez:
this is by default a massive improvement to bays magic system and will be used as a framework for further systems involving reality bending/thaumaturgy.

PR #1563, opened 2024-03-21
Keybind for locking Face Direction

Saicchi:
## About the Pull Request

Add a keybind 'F' to lock facing direction.

## Why It's Good For The Game

Mouse clicking or typing out verbs for an often used command is not cool.

## Changelog

:cl:
add: Modifiable keybind 'F' to lock facing direction
/:cl:

### DIFF
--- a/code/modules/keybindings/mob.dm
+++ b/code/modules/keybindings/mob.dm
@@ -118,6 +118,17 @@
 	M.stop_pulling()
 	return TRUE
 
+/datum/keybinding/mob/lock_facing_direction
+	hotkey_keys = list("F")
+	name = "lock_facing_direction"
+	full_name = "Face Direction"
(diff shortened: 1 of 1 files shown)


TerryTehCoder:
I'm all for this, this is often needed and difficult to access. One thing I should add is that I've seen HUD buttons implemented for this in other codebases. It might be worth considering snatching one of those for this as well given how important the function is here.

Saicchi:
> I'm all for this, this is often needed and difficult to access. One thing I should add is that I've seen HUD buttons implemented for this in other codebases. It might be worth considering snatching one of those for this as well given how important the function is here.

Human players already have a HUD button to lock their facing direction. Robots don't need one because they can see 360 degrees.

PR #1571, opened 2024-03-24
Fix map config path

checkraisefold:
## About the Pull Request

Fixes the config path so the map year and other stuff can be configured

## Why It's Good For The Game

This is supposed to be set and is set on the torch on tegu

## Changelog

:cl:
code: Site 53 map config path fixed.
/:cl:


checkraisefold:
Tested and works with #1577

Tsurupeta (approved):

PR #1576, opened 2024-03-25
Make show player info get age correctly

checkraisefold:
## About the Pull Request

This might fix player age: unknown issue for people who are connected to the server
Can't test due to it requiring a DB and I am too lazy to set up maria locally but this is better logic anyway

## Why It's Good For The Game

No player age unknown shenanigans

## Changelog

:cl:
admin: Show Player Info verb/notes should no longer show player age 'unknown' for connected clients
/:cl:

### DIFF
--- a/code/modules/admin/admin.dm
+++ b/code/modules/admin/admin.dm
@@ -84,7 +84,6 @@ var/global/floorIsLava = 0
 
 
 /datum/admins/proc/show_player_info(key as text)
-
 	set category = "Admin"
 	set name = "Show Player Info"
 	if (!istype(src,/datum/admins))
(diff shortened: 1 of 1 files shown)


code/modules/admin/admin.dm
```diff
@@ -96,8 +95,9 @@ var/global/floorIsLava = 0
 	var/list/dat = list()
 
 	var/p_age
+	var/key_c = ckey(key)
```
Tsurupeta, line 98:
```suggestion
```

code/modules/admin/admin.dm
```diff
@@ -96,8 +95,9 @@ var/global/floorIsLava = 0
...
 	for(var/client/C in GLOB.clients)
-		if(C.ckey == key)
+		if(C.ckey == ckey(key_c))
```
Tsurupeta:
```suggestion
		if(C.ckey == ckey(key))
```

Tsurupeta (changes requested):
???

code/modules/admin/admin.dm
checkraisefold: Calling a string transformation for every client on the entire server seems slow unless byond compiler optimizes this out which i doubt

Tsurupeta (approved):

PR #1577, opened 2024-03-25
Change using map config load order in mc

checkraisefold:
## About the Pull Request

Fix ancient bug that even exists on tegu i think that means the map config just doesnt load

## Why It's Good For The Game

Map config works
#1571 does something

## Changelog

:cl:
fix: Map config actually loads now
/:cl:


checkraisefold:
Tested and working

Tsurupeta (approved):

PR #1578, opened 2024-03-25
Adds override for timelocks for roles already played

TheDarkElites:
## About the Pull Request

Allows timelocks to be overridden by the role itself to allow for individuals who stacked hours when timelocks were disabled to play the roles they put time into.

## Why It's Good For The Game

Makes people happy, makes me tired

## Changelog

:cl:
code: Timelocks will now also accept the time spent in the currently checked role for time reqs.
/:cl:


TheDarkElites:
since this is a temporary fix this should only be TM'D until moving pains are over

additionally, due to #1583 this might no longer even be necessary

PR #1583, opened 2024-03-25
Change almost all timelocks to be department based

checkraisefold:
## About the Pull Request

This is a rework to have time requirements be almost entirely department based, and also tune down the restrictions to be more reasonable. This will definitely need some input to make right, though.

## Why It's Good For The Game

The extremely overwhelming player complaints whenever timelocks are turned on is literally not being able to play any jobs. I, as a Trial GM, do not have R_TIMELOCK permissions (to bypass timelocks) and with a few hundred hours on the server can't access almost any jobs.
This is a consequence of most timelocks requiring **SPECIFIC JOB** restrictions, along with being too restrictive in themselves.

Timelocks are unsustainable to have on with the current requirements and dramatically reduce pop. Let's see if we can fix that.

## Changelog

:cl:
balance: Timelock requirements for all jobs have been significantly re-tuned.
/:cl:

### DIFF
--- a/code/game/jobs/job/command.dm
+++ b/code/game/jobs/job/command.dm
@@ -20,7 +20,7 @@
 	access = list()				// see get_access() override
 	minimal_access = list()		// see get_access() override
 
-	requirements = list(EXP_TYPE_COMMAND = 1800)
+	requirements = list(EXP_TYPE_COMMAND = 1200)
 
 /datum/job/captain/get_access()
 	return get_all_site_access()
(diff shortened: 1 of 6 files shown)


TheDarkElites:
Timelocks were already significantly retuned to be much smaller and now you want to make them even smaller? And why? There is already a PR in progress to make it so that any time you spent in a role can be counted for the time reqs of that role, allowing for people who spent hours playing RD or CMO to be able to play that role once timelocks are enabled. In my opinion, this is overzealous and quick judgment without even giving the current settings any actual playtime to see how it affects gameplay.

Not to mention, time requirements were not solely department-based **Because the separation of roles for most departments is by seniority, not specialization**, I.E. I shouldn't be able to play senior researcher from only playing junior researcher.

TheDarkElites:
Currently working with @checkraisefold to get this more in line with expected standards

TheDarkElites (approved):
better

TheDarkElites:
Happy with tm phase, merging

PR #1584, opened 2024-03-25
Urgent 173 invisible gibs fix

checkraisefold:
## About the Pull Request

173 can breach like 2 minutes after being cleaned because of invisible gibs under the floor
See responsible code that doesnt delete the gibs and instead does stupid bullshit:
https://github.com/Foundation-19/Foundation-19/blob/dev/code/game/objects/effects/decals/Cleanable/humans.dm#L37

## Why It's Good For The Game

no more insta breach

## Changelog

:cl:
fix: 173 dirt in containment chamber is counted correctly
/:cl:

### DIFF
--- a/code/modules/SCP/SCPs/SCP-173.dm
+++ b/code/modules/SCP/SCPs/SCP-173.dm
@@ -359,6 +359,10 @@
 		return 0
 	var/feces_amount = 0
 	for(var/obj/O in A)
+		if(istype(O, /obj/effect/decal/cleanable/blood))
+			var/obj/effect/decal/cleanable/blood/B = O
+			if(B.amount == 0)
+				continue
 		if(O.type in defecation_types)
 			feces_amount += 1
 			continue


checkraisefold:
dont know if the TM is working on prod since host refuses to give me a role with +debug for a period of 5 seconds to proccall CheckFeces

TheDarkElites:
> dont know if the TM is working on prod since host refuses to give me a role with +debug for a period of 5 seconds to proccall CheckFeces

There is this cool thing you can do where you ask me to check

TheDarkElites:
Really we should be fixing the code that causes invisible gibs

checkraisefold:
> Really we should be fixing the code that causes invisible gibs

This is true but I think its like a thing to stack gibs and reuse them later? idfk

Tsurupeta (approved):

Tsurupeta:
make an issue to fix the invisible gibs later, not merging it until issue is made so we wont forget about it (we will still forget about it)

PR #1586, opened 2024-03-26
Light replacer use in hand to replace lamps in area

Saicchi:
## About the Pull Request

Lets the light replacer replace all broken adjacent lights after a 1 second do_after

## Why It's Good For The Game

Clicking the tiny pixels of a broken light fixture is hard.

## Changelog

:cl:
add: light replacer replace lights in area
/:cl:

### DIFF
--- a/code/game/objects/items/devices/lightreplacer.dm
+++ b/code/game/objects/items/devices/lightreplacer.dm
@@ -116,15 +116,17 @@
 			return
 
 /obj/item/device/lightreplacer/attack_self(mob/user)
-	/* // This would probably be a bit OP. If you want it though, uncomment the code.
-	if(isrobot(user))
-		var/mob/living/silicon/robot/R = user
-		if(R.emagged)
-			src.Emag()
-			to_chat(usr, "You shortcircuit the [src].")
+	if(uses <= 0 || !isliving(user))
+		return
+	if(!do_after(user, 0.5 SECONDS, src))
+		return
+	add_fingerprint(user)
+	for(var/obj/machinery/light/O in range(1,user))
+		if(O.get_status() == LIGHT_OK)
+			continue
+		if(uses <= 0 || !do_after(user, 0.2 SECONDS, src))
 			return
-	*/
-	to_chat(usr, "It has [uses] lights remaining.")
+		ReplaceLight(O, user)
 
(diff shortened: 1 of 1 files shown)


code/game/objects/items/devices/lightreplacer.dm
```diff
@@ -116,15 +116,16 @@
...
 /obj/item/device/lightreplacer/attack_self(mob/user)
-	/* // This would probably be a bit OP. If you want it though, uncomment the code.
-	if(isrobot(user))
-		var/mob/living/silicon/robot/R = user
-		if(R.emagged)
-			src.Emag()
-			to_chat(usr, "You shortcircuit the [src].")
-			return
-	*/
-	to_chat(usr, "It has [uses] lights remaining.")
```
checkraisefold, line 127: ?

Saicchi: The user can see the remaining lights by examining the tool.

checkraisefold: > The user can see the remaining lights by examining the tool.

oh this is fine then

code/game/objects/items/devices/lightreplacer.dm
```diff
@@ -116,15 +116,16 @@
...
+	if(!isliving(user) || !do_after(user, 1 SECONDS, src))
+		return
+
+	for(var/obj/machinery/light/O in range(1,user))
+		if (!CanUse(user))
```
TheDarkElites: This should just be an early return outside of the loop instead of a break statement inside

TheDarkElites (changes requested):

code/game/objects/items/devices/lightreplacer.dm
Saicchi: Changed it.

TheDarkElites: When I say early return I mean having the return statement **outside of the for loop**

Saicchi: Rewrote it a bit to move the check to outside the loop and give a small delay when replacing each light. Does it look better now?

TheDarkElites (approved):

PR #1600, opened 2024-03-28
AI and Robot lights keybind

Saicchi:
## About the Pull Request

Keybinding for Robots to toggle their lights and AI to toggle camera light.

## Why It's Good For The Game

QoL is good.

## Changelog

:cl:
add: keybind 'V' for robot and AI lights
/:cl:

### DIFF
--- a/code/modules/keybindings/robot.dm
+++ b/code/modules/keybindings/robot.dm
@@ -1,3 +1,20 @@
+/datum/keybinding/silicon
+	category = CATEGORY_ROBOT
+
+/datum/keybinding/silicon/toggle_lights
+	hotkey_keys = list("V")
+	name = "toggle_lights"
+	full_name = "Toggle Lights"
+	description = "Toggle camera lights"
+
+/datum/keybinding/silicon/toggle_lights/down(client/user)
+	if(isrobot(user.mob))
+		var/mob/living/silicon/robot/S = user.mob
+		S.toggle_lights()
+	if(isAI(user.mob))
+		var/mob/living/silicon/ai/S = user.mob
+		S.toggle_camera_light()
+
 /datum/keybinding/robot
 	category = CATEGORY_ROBOT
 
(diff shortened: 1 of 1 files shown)


code/modules/keybindings/ai.dm
TheDarkElites, line 1: Ai doesn't need its own file, just have one proc for both robot and AI and then just checks is the user is an AI or a normal borg

code/modules/keybindings/robot.dm
```diff
@@ -70,3 +70,13 @@
...
 		R.uneq_active()
 	return TRUE
+
+/datum/keybinding/robot/toggle_lights
+	hotkey_keys = list("V")
+	name = "toggle_lights"
+	full_name = "Toggle Lights"
+	description = "Toggle integrated lights"
+
+/datum/keybinding/robot/toggle_lights/down(client/user)
```
TheDarkElites: these should have istype checks

TheDarkElites (changes requested):

code/modules/keybindings/robot.dm
Saicchi: How about now?

code/modules/keybindings/robot.dm
```diff
@@ -1,3 +1,20 @@
...
+/datum/keybinding/silicon/toggle_lights
+	hotkey_keys = list("V")
+	name = "toggle_lights"
+	full_name = "Toggle Lights"
+	description = "Toggle camera lights"
+
+/datum/keybinding/silicon/toggle_lights/down(client/user)
+	if(isrobot(user.mob))
+		var/mob/living/silicon/robot/S = user.mob
+		S.toggle_lights()
```
TheDarkElites, line 13: This probably won't be a problem, but I would rather you have this as an if and then if else rather than two ifs. This shouldn't happen, but you never know if both statements could run.

TheDarkElites (dismissed):

PR #1619, opened 2024-03-30
Fix goal system

checkraisefold:
## About the Pull Request

1. fix runtime caused by using QDEL_LIST_NULL on an associative list
2. fix goal list randomly disappearing; was noticeable by staff because of aghosting

alternate fix for the goal list disappearing is to change the duplication mode of the goal container to unique, so that it persists across mind inits

## Why It's Good For The Game

fix John Pizzacode

## Changelog

:cl:
fix: Fix runtime in goal container code.
fix: Fix goal list breaking completely when re-entering corpse while ghosted. Relevant for admins who aghost.
/:cl:


checkraisefold:
tested & working

Tsurupeta (approved):

PR #1630, opened 2024-04-01
Rags arent glass silly

tichys:
Will have to fix the bucket later


TheDarkElites (approved):
ANOTHER FIX TO OUR HORRID PATHING? OH MY GOD ITS A MIRACLE

PR #1649, opened 2024-04-09
Map renderer hide_invisible set

aidan-bowman:
## About the Pull Request

Muh webmap

## Changelog

Nothing player facing.

### DIFF
--- a/SpacemanDMM.toml
+++ b/SpacemanDMM.toml
@@ -4,7 +4,7 @@ dreamchecker = true
 [map_renderer]
 hide_invisible = [
 	"/obj/effect/landmark",
-    "/turf/unsimulated/mineral",
+	"/turf/unsimulated/mineral",
 ]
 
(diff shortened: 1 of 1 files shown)


SpacemanDMM.toml
```diff
@@ -1,6 +1,12 @@
 [langserver]
 dreamchecker = true
 
+[map_renderer]
+hide_invisible = [
+	"/obj/effect/landmark",
+    "/turf/unsimulated/mineral",
+]
+
```
AffectedArc07:
```suggestion
[map_renderer]
hide_invisible = [
	"/obj/effect/landmark",
	"/turf/unsimulated/mineral",
]

```

AffectedArc07:
For those asking why this is important
image

I am not hosting a webmap where each layer is 100mb, that will take people ages to download and the map wont be usable.

PR #1698, opened 2024-06-27
Self-Destruct authentication updates

TerryTehCoder:
Updates keycard authentication devices to ACCESS_KEYCARD_AUTH, which, in hindsight you would think would make more sense than Level-4 Science and Level-4 Engineering.

Also gave all Heads of Staff level-1 engineering access (if they didn't already have it)

Heads of staff can now actually use their Keycard Authenticators and access the self-destruct system located in Engineering.

## Why It's Good For The Game

Heads of staff were unable to use their own keycard authenticators which was silly, and I have no idea how long this has been the case for. I actually just wanted to add a Self-Destruct chapter for the Site-Directors wiki page so I could make an Aliens Reference, but then I got caught up in this Rabbit Hole and figured eh I might as well clean it up.

There is still very little risk to Heads of Staff having access to the Self-Destruct room asides from the obvious that it is more likely someone will attempt to antagonise them for their keycard. Alternatively the Traitor ID Access code program has existed for Eons and while it does take a while you only need One Code to get in. The Nuclear Authentication Disk is still necessary to even begin S2 (Code Authentication) which requires not one but Two head ID's.

## Changelog

:cl:
Changed: Keycard Authenticator access requirement from Level-4 Engineering and Level-4 Science to ACCESS_KEYCARD_AUTH
Added: Level-1 Engineering access to Chief Medical Officer and Assistant Medical Director.
Updated: Airlock access requirements for Self-Destruct room are now based on Keycard Authenticaton access rather than Admin Level 5 which no one but the site director had.
/:cl:


TerryTehCoder:
Thank you for catching the missing comma.

PR #1705, opened 2024-07-01
Raises Psionic latency so the Psionics Division of research can actually work

Joel-682:
## About the Pull Request
Raises the latency chance from 2 to 5. This means for every 20 pop, there's ONE psionic on average.

## Why It's Good For The Game

This allows the Psionics division to have more work more often and maybe not be pointless. I would even suggest raising it to 10, so that Psionics has stuff to do regularly

## Changelog

:cl:

config: Raised the  var/psi_latency_chance from 2 to 5. 

/:cl:

### DIFF
--- a/code/game/jobs/job/job.dm
+++ b/code/game/jobs/job/job.dm
@@ -76,7 +76,7 @@
 	/// Starting psi faculties, if any.
 	var/list/psi_faculties
 	/// Chance of an additional psi latency, if any.
-	var/psi_latency_chance = 2
+	var/psi_latency_chance = 5
 	/// If psionic, will be implanted for control.
 	var/give_psionic_implant_on_join = FALSE
 


WoodenTucker:
CALL THEM LATENT THAMUOLOGISTS PLEAAAAAAAAAAAAAAAAAAASE

Joel-682:
Fuck you **Atomiscizes your idea or however it's spelt**

Regardless, what do you mean? Changing the instances of psionic/psy mentions into "Thaumologics/Thaumology"? Say, in the Psy-assayer for example?

WoodenTucker:
im apart of the LORE team... you do NOT atomize my idea.. .i make yuor character gay.


i personally view 'psionics' as thaumatology and wizard spell shit as psionics, to fix  this i just want them appropriately renamed w/ references, like the psy-assayer yeah

Joel-682:
I'll take a crack at it sure, how hard can it be.

checkraisefold:
yeah no 2% is fair at any pop above like 20
how about instead just make the chance player count based. guarantee at least one psionic player per round if the player count is greater than 20

Joel-682:
I- Don't think so? There's a couple of roles related to psionics, and they get to do nothing but make chemicals related to psionics that don't get used because no psionics are found. And again, with this setting on, whenever we hit **100** pop, we get FIVE whole psionics users. And they still need to be found. You got to remember that, psionic latency != psionic powers. Like, out of those five, at least one will go unfound (Either by death, or lack of searching) and it's most likely more than that. 



I'm not comfortable enough doing any good code with formulas so I won't take a crack at doing a player based chance.

PR #1709, opened 2024-07-05
fix orbit menu bugs

checkraisefold:
## About the Pull Request

fixes the global list pollution that was probably breaking the performance along with other shit on the entire server
fixes the cancel button sending you to hell

## Why It's Good For The Game

common lest code W

## Changelog

:cl:
fix: Orbit menu cancel button now works correctly
fix: Orbit menu is no longer extremely long and no longer duplicates its items when opened
/:cl:


checkraisefold:
tested & working
the new `list()` every time I thought could pose memory problems but `tgui_input_list` garbage collects correctly so its fine

PR #1712, opened 2024-07-05
More R-UST fixes

checkraisefold:
## About the Pull Request

you can now actually build a full rust midround technically. before it was impossible/very shitty because you could not set the tags of fuel injectors so they would never be able to show up on the computer for them

also the pump is no longer poorly placed in the control room which was making noise every round forever

## Why It's Good For The Game

i love engineering

## Changelog

:cl:
fix: Fuel injectors can have their tag set with a multitool, allowing full R-UST setup building/restoration midround
fix: Pump in R-UST control room has been moved to chamber itself, no longer creating air leak noise
/:cl:


checkraisefold:
tested & working

PR #1744, opened 2024-07-10
Moves Self-Destruct to Administration

ChromeFoxxity:
## About the Pull Request

Deleted the silo from Engineering, moving it to a more secure area in Administration.

## Why It's Good For The Game

More security.

## Changelog

:cl:
add: New Silo.
del: Old Silo.
/:cl:

### DIFF
--- a/code/modules/events/prison_break.dm
+++ b/code/modules/events/prison_break.dm
@@ -74,9 +74,6 @@
 	areaType = list(/area/site53/engineering)
 	areaNotType = list(
 		/area/site53/engineering/maintenance,
-		/area/site53/engineering/selfdestruct,
-		/area/site53/engineering/lowernukeladders,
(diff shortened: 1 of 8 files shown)


checkraisefold:
i dont like this the entire point is its out of the way and hard to get to, not just "site director walks outside his office and can hit da nuke"
the entire point in-character as well is that it is at the lowest point of the site. you factually do not want it higher up so you don't make the surface completely uninhabitable for 200 years, this puts it way higher up and doesnt make sense in roleplay even with improved nukes in ingame year

checkraisefold:
this has been denied for game balancing
the SD/other heads should not be able to walk out of their office/out of the panic bunker and walk directly into the nuke room as soon as an SCP breaches with no difficulty. it's intended to be a Final Option and also not be super easy to use, that's why it's in engineering

PR #1774, opened 2024-07-20
THE mech update.

MLGTASTICa:
## About the Pull Request

Adds a plethora of features to mech , namely
1)Reworked mech's to be able to draw from multiple power sources , namely , their power button was replaced with a power hardpoint (special icon pending)
2)Added backup-power to mechs ,this is also implemented as a hardpoint that will be drawn from when no other source exists , it currently can only supply arms & equipment , mechs using only auxiliary power are not able to move.
3)Added the mech engine (runs off welding fuel, has a small 500 units cell, and 200 units of reagent space)
4)Added the mender (ported from bay , a exosuit-mender that applies medigel over wounded areas on someone)
5)Added a alternative mode to exosuit floodlight (directed , which puts the light source further away, focusing it)
6)Added more chemicals to the exosuit medical sleeper
7)Added in exosuit strafing(with support  to toggle by pressing F)

Fixes:
1)Fixed mech shields overlaying incorrectly
2)Fixed mech NV
3)Fixed a edge-case for installing modules.
4)Fixed mech FOV

Other changes
1)All mech equipment no longer requires the canopy to be lowered to use
2)Added frameworks for mech_flags , mech_equipment flags , and propulsion flags, adjusting code to support their functionality
3)Mech power was reworked.

## Why It's Good For The Game
Mechs cool

## Changelog

:cl:
refactor: Reworked mech power , now mechs draw power from power hardpoints , which are colored in yellow.
add: Added the auxiliary power cell slot to combat mechs , this will be used for weapons & arm actions when the main one is not available
add: Added the mech engine , a alternative to the power cell,the engine has a internal cell of just 500 units ,but will use welding fuel to recharge itself slowly
add: Mech mender, used to heal pacients or scan them
balance: buffed the mech sleeper to include more chemicals
add: added in exosuit strafing
add: added a new mode to the exosuit floodlight , DIRECTED which will cause it to light in a cone ahead of the exosuit.
fix: Fixed mech shields not properly layering
fix : Fixed mech NV
/:cl:


checkraisefold:
workflows approved

checkraisefold:
workflows approved again. god I don't know why it doesnt mark you as approved forever the first time I approve it

tools/CreditsTool/obj/Debug/.NETFramework,Version=v4.6.1.AssemblyAttributes.cs
checkraisefold, line 1: ?

tools/CreditsTool/obj/Debug/CreditsTool.csproj.AssemblyReference.cache
checkraisefold, line 1: ?

checkraisefold (changes requested):

tools/CreditsTool/obj/Debug/.NETFramework,Version=v4.6.1.AssemblyAttributes.cs
MLGTASTICa: Autogenerated project files , VSC does this for every C++ subproject file. Removed em now.

tools/CreditsTool/obj/Debug/CreditsTool.csproj.AssemblyReference.cache
MLGTASTICa: Autogenerated project files , VSC does this for every C++ subproject file

code/modules/mechs/equipment/utility.dm
```diff
@@ -205,19 +208,106 @@
...
+	if(mode == MODE_CIRCULAR || !on)
+		return
+	var/turf/current = get_turf(owner)
+	var/turf/temp = get_step(src, owner.dir)
+	for(var/i = 0, i <= effect_steps, i++ )
+		if(temp.opacity)
+			break
+		for(var/atom/thing in temp.contents)
+			if(thing.opacity)
+				goto end
```
checkraisefold: i dont like this change it to a flag with an if check
this is also extremely expensive and you should figure out how to change it anyway

test/check-paths.sh
checkraisefold, line 1: prior comment

checkraisefold (changes requested):

code/modules/mechs/equipment/utility.dm
MLGTASTICa: propose a better implementation?

test/check-paths.sh
MLGTASTICa: goto is best in this case, the other alternative is to litteraly have a var that if true, to break the outer loop too.

MLGTASTICa:
Converted back to draft due to a few adjustments being needed

code/modules/mechs/equipment/utility.dm
ghost: The way you have changed canopies has probably borked the code to interact with items in the mech, check that you didn't actually break this before this is merged

PR #1780, opened 2024-07-23
Reworks slime handling.

MLGTASTICa:
## About the Pull Request
Adds a fast_mobs subsystem and puts slimes under it
removes the replicating handle_ai calls
I saw this during the mech PR and i couldn't cope with such lag existing duty-free

## Why It's Good For The Game
Fix good , doesn't crash the server if none of the stop conditions are fulfilled before the infinite loop begins.
## Changelog

:cl:
refactor: Refactored slime AI handling.
/:cl:


checkraisefold:
I really don't like the idea of a fast mobs subsystem

MLGTASTICa:
> I really don't like the idea of a fast mobs subsystem



> I really don't like the idea of a fast mobs subsystem

You'd rather prefer that we simulate a subsystem by creating looping handle_ai calls with 1 tick delays everytime handle_regular_ai is called ? (2 seconds). This is the reason the server sometimes just crashes up or starts lagging massively, the infinite handle_ai loop just builds up in cases where none of the stop condition are valid.

MLGTASTICa:
> > I really don't like the idea of a fast mobs subsystem

You'd rather prefer that we simulate a subsystem by creating looping handle_ai calls with 1 tick delays everytime handle_regular_ai is called ? (2 seconds). This is the reason the server sometimes just crashes up or starts lagging massively, the infinite handle_ai loop just builds up in cases where none of the stop condition are valid.
This regularly bricks the timer subsystem on local servers.

PR #1783, opened 2024-07-26
SCP-458

tichys:
Added the infinite pizza box


ChuChu42:
suggwstion: instead of it being only cheese pizza take the infinite pizza box code from a server like TG so it gives you your "favorite pizza" which changes for each person

that way it has more researchability

PR #1800, opened 2024-09-11
Converts S53 to One File, cleans up old files.

TerryTehCoder:
## About the Pull Request

**Converted S53 into one Multi-Z file instead of five different Z levels**. Moved 106 PD to z1_admin. Got rid of old files that are no longer used or were never used but were being stored in the map folder.

Did you know 106's PD had its own Z level with Nothing Else on it, and it took up roughly 33x20 tiles worth of that 255x255?

The CI base is already functional on the admin-z, so the file is either a duplicate or not used at all.

-Removed admin_levels from define (handeled in a separate file, not the main map file) 

-Removed empty_levels from define (unused entirely) 

-Removed accessible_z_levels from define. (You should not be able to access any of the outer map edges **irregardless** due to mapping design, but there is no reason we should have percentile chances for drifting onto other Z-levels, we're not in space.) 

-Added sealed_levels define for all player-z's which prevents players from stepping onto the edge of the map even if they did manage to somehow navigate past unbreakable rock.

## Why It's Good For The Game

Ease of modifiability, having everything in one file is marginally easier to toy with and visualize than everything in four separate files. Old mapping files while archaic and fun to look at do not need to be stored in the repository.

This also means Multi-Z rendering in the editor which is nice to have.
image

## Changelog

:cl:
Modified: Site53 is now a single multi-z file rather than five different Z-levels.
Added: Sealed_Levels define to all player levels
Removed: Admin_Levels and Empty_Levels from S53 define. Admin_Z is handeled in its own file and no empty levels exist in S53.
/:cl:

Functional during self-testing, could benefit from further review while the branch is up.

### DIFF
--- a/code/datums/mind/mind.dm
+++ b/code/datums/mind/mind.dm
@@ -61,7 +61,7 @@
 
 	var/list/initial_email_login = list("login" = "", "password" = "")
 
-
+	var/ambitions
 
 /datum/mind/New(key)
 	src.key = key
@@ -128,6 +128,7 @@
 	else
 		out += "None."
 	out += "<br><a href='?src=\ref[src];obj_add=1'>\[add\]</a><br><br>"
+	out += "<b>Ambitions:</b> [ambitions ? ambitions : "None"] <a href='?src=\ref[src];amb_edit=\ref[src]'>\[edit\]</a></br>"
 
 	show_browser(usr, out, "window=edit_memory[src]")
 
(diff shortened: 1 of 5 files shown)


TerryTehCoder:
Will probably have to redo it when some of the older mapping PR's go through but it's notoriously easy to implement, so I don't mind.

checkraisefold (approved):
I was going to do this myself thank god someone's finally done it
Let's DNM this until some of the bigger mapping prs that have been sat there get merged, we should speed them up now
The 106 Z level was especially awful and terrible for memory usage thanks for tossing it

TerryTehCoder:
This might actually break nanomap, I didn't consider how that code recognizes map images for single file multi-Z maps. Honestly Nanomap in general could use a rework since having hard-coded image storage causes issues whenever the map is updated and nanomap is left behind.

checkraisefold:
> This might actually break nanomap, I didn't consider how that code recognizes map images for single file multi-Z maps. Honestly Nanomap in general could use a rework since having hard-coded image storage causes issues whenever the map is updated and nanomap is left behind.

no one uses nanomap we can just fix it later

checkraisefold (approved):
All blocker PRs merged, good to resolve conflicts/re-merge the map file and push this through

PR #1802, opened 2024-09-16
Allows shuttle landing and takeoff audio to be played in both current and destination.

TerryTehCoder:
## About the Pull Request

Adds a variable to shuttle.dm which allows shuttle landing and takeoff audio to be played at the same time in both the current and destination location. 

## Why It's Good For The Game

Shuttles without transitions previously failed to play landing audio, this resolves that in its specific usecase. It's nice to have for things such as small Lifts where you want the audio to signify its movement in both the takeoff zone and the landing zone.

## Changelog

:cl:
Added: Shuttle audio variable which makes the shuttle play its takeoff and landing audio in both current and destination areas at the same time during takeoff and landing.
/:cl:


TerryTehCoder:
To clarify this does not mean takeoff and landing audio play at the same time, it means takeoff will play in both areas at the same time, and landing will play in both areas at the same time. Although this isn't Super relevant for us as most of our shuttles use the same takeoff and landing audio clip.

code/modules/shuttles/shuttle.dm
```diff
@@ -19,6 +19,8 @@
 	var/sound_takeoff = 'sounds/effects/shuttle_takeoff.ogg'
 	var/sound_landing = 'sounds/effects/shuttle_landing.ogg'
 
+	var/audioalways = 0 //Do we want the audio to play in the destination and takeoff location? Useful for things such as lifts with no transition location to trigger landing audio.
```
checkraisefold:
```suggestion
	var/audioalways = FALSE //Do we want the audio to play in the destination and takeoff location? Useful for things such as lifts with no transition location to trigger landing audio.
```

code/modules/shuttles/shuttle.dm
```diff
@@ -126,7 +132,11 @@
 			while (world.time < arrive_time)
 				if(!fwooshed && (arrive_time - world.time) < 100)
 					fwooshed = 1
-					playsound(destination, sound_landing, 100)
+					if(audioalways)
+						playsound(current_location, sound_landing, 100)
+						playsound(destination, sound_landing, 100)
+					else
+						playsound(destination, sound_landing, 100)
```
checkraisefold: If it's for lifts with no transition then why is this in long_jump which always has an interim/transition

checkraisefold (changes requested):
I don't see the point of this PR. The "lift with no transition location" example does not make sense in my mind

TerryTehCoder:
> I don't see the point of this PR. The "lift with no transition location" example does not make sense in my mind

Sometimes we want to exploit shuttles in unconventional ways built into the structure of maps. In such occasions you want the audio to be heard at both a takeoff and landing location. I could give you an example if you'd like. 

Something similar to a structure that can be raised or lowered  between Z's requires this change to allow the audio to be heard in both Z's. I can clarify with a video example if this is still confusing language.

checkraisefold:
> > I don't see the point of this PR. The "lift with no transition location" example does not make sense in my mind
> 
> Sometimes we want to exploit shuttles in unconventional ways built into the structure of maps. In such occasions you want the audio to be heard at both a takeoff and landing location. I could give you an example if you'd like.
> 
> Something similar to a structure that can be raised or lowered between Z's requires this change to allow the audio to be heard in both Z's. I can clarify with a video example if this is still confusing language.

please toss a video example

TerryTehCoder:
> > > I don't see the point of this PR. The "lift with no transition location" example does not make sense in my mind
> > 
> > 
> > Sometimes we want to exploit shuttles in unconventional ways built into the structure of maps. In such occasions you want the audio to be heard at both a takeoff and landing location. I could give you an example if you'd like.
> > Something similar to a structure that can be raised or lowered between Z's requires this change to allow the audio to be heard in both Z's. I can clarify with a video example if this is still confusing language.
> 
> please toss a video example



https://github.com/user-attachments/assets/96de5f03-45a7-4489-b182-6953c9ce5ac2

code/modules/shuttles/shuttle.dm
TerryTehCoder: This is actually an oversight, and unrelated. Good catch. 

Though it may be the case that something like this is used in the future it's out of the scope of the PR, so I'll gut it from Long Jump.

checkraisefold (changes requested):
i am willing to merge this once you do the review comment

TerryTehCoder:
> i am willing to merge this once you do the review comment

Review is resolved.

checkraisefold (approved):

PR #1810, opened 2024-10-07
Multizsite53 - #2

TerryTehCoder:
## About the Pull Request

**Converted S53 into one Multi-Z file instead of five different Z levels**. Moved 106 PD to z1_admin. Got rid of old files that are no longer used or were never used but were being stored in the map folder.

Did you know 106's PD had its own Z level with Nothing Else on it, and it took up roughly 33x20 tiles worth of that 255x255?

The CI base is already functional on the admin-z, so the file is either a duplicate or not used at all.

-Removed admin_levels from define (handeled in a separate file, not the main map file) 

-Removed empty_levels from define (unused entirely) 

-Removed accessible_z_levels from define. (You should not be able to access any of the outer map edges **irregardless** due to mapping design, but there is no reason we should have percentile chances for drifting onto other Z-levels, we're not in space.) 

-Added sealed_levels define for all player-z's which prevents players from stepping onto the edge of the map even if they did manage to somehow navigate past unbreakable rock.

## Why It's Good For The Game

Ease of modifiability, having everything in one file is marginally easier to toy with and visualize than everything in four separate files. Old mapping files while archaic and fun to look at do not need to be stored in the repository.

This also means Multi-Z rendering in the editor which is nice to have.
image

## Changelog

:cl:
Modified: Site53 is now a single multi-z file rather than five different Z-levels.
Added: Sealed_Levels define to all player levels
Removed: Admin_Levels and Empty_Levels from S53 define. Admin_Z is handeled in its own file and no empty levels exist in S53.
/:cl:

Functional during self-testing.


checkraisefold (approved):
Undergoing a test merge to make sure nothing Fucking Explodes

checkraisefold:
test merge has gone fine

PR #1821, opened 2024-10-11
Mountains in inappropriate places Fix

AlanParkerMasters:
## About the Pull Request
Replaces mountain tiles in certain areas inside of the facility that do not fit where they should be. Some of these tiles had light tubes in them. 

## Why It's Good For The Game
Makes the facility look cleaner. The facility, which is surrounded by built structures, shouldn't have strange walls of mountain inside of it. 

## Changelog

:cl:
Replaced the mountain tiles in Medbay that held plants/fauna with actual walls since the entire facility follows this aesthetic.
Also replaced a random section of mountain that was just vibing in maintenance with some fitting walls.
/:cl:


checkraisefold (changes requested):
Please scan the upper and lower Z-levels to make sure this doesnt cause a instant 106 breach or allow 106 access to areas it normally would not be able to get into
Pretty sure he can go into the non-mountain walls

AlanParkerMasters:
Checked and spoke with Check, the edited walls shouldn't interfere or cause any problems with neighboring Z-levels

checkraisefold (approved):

PR #1832, opened 2024-11-11
Update command.dm

islaughterfunny:
Gives the reps MTF access (you wont make it there in a regular round anyways)


checkraisefold:
No

